PR #7, opened 2023-04-18
Set up typechain

Hyodar:
Simple typechain setup. Currently only `IERC20`, `EigenPodManager` and `StrategyManager` types, as these are the ones that should be used more frequently, but more can be added on the go. Ideally we should get the `eignlayr-contracts` project as a package, compile it and get all the types on the go (no need to have ugly ABIs on the repo), but I don't think this is the focus for this draft.

Adding this because it's way neater than not having the types for indexing.


EdgarBarrantes (approved):

abi/EigenPodManager.json
swapnilraj, line 1: Should we have a submodule for the eigenlayer contracts and generate the abi in the build step

Hyodar: Yes, also mentioned this in the PR description. IMO this can even be in the postinstall script. I think the options here are either making a script with solcjs or just using forge straight away - is it too much of an issue to consider a forge installation on the server? If not, this is really easy.

Hyodar: Plus, no need for submodules, we can get it as an npm package through git even though it's not published in the npm registry.

Hyodar: Well, actually submodules are a lot better because `npm` doesn't get the `.git` when fetching a git package. Added the contracts repo as a submodule and removed the abis on [c59435a](https://github.com/NethermindEth/restaking-dashboard/pull/7/commits/c59435a2110639781737ec6bebbe0208ad79a9ba). Still, I made a separate command for vercel build (`vercel-build`) to take these steps (submodule update, forge installation, typing generation) into account, so we'll need to change it to that to make deployment work.

PR #10, opened 2023-04-19
Add indexing utilities

Hyodar:


abi/EigenPodManager.json
swapnilraj, line 1: Should we make this a submodule generate the abi's in the build process?

Hyodar:
I think this is fully ready for review. The only missing part atm, I'd say, is removing the leading underscore from the table names both here and on Supabase, which was added as a way of avoiding conflict with existing table schemas. However, I'd like to do this when everyone's on so we're sure we're not breaking anything and any views can be updated.

lib/indexing/utils/beaconProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+  const req = await fetch(
+    `${BEACON_PROVIDER_URL}/eth/v1/beacon/states/${state}/validators?${encodeIndexes(indexes)}`
+  );
+  const json = (await req.json()) as BeaconApiResponse<ValidatorResponse>;
+
+  return json.data.map(validatorResp => ({
+    index: parseInt(validatorResp.index),
+    balance: BigInt(validatorResp.balance),
+    status: validatorResp.status,
+    validator: {
```
swapnilraj, line 83: you can use the spread operator here

swapnilraj:
Very readable code, thank you!
The calls that use a provider don't handle cases where it fails.
How likely is it that they might fail? Should we connect to our goerli node to minimize that?

lib/indexing/utils/beaconProvider.ts
Hyodar: Yup! I didn't use it because I thought I was overriding most keys anyway, but I agree it looks a bit more compact. On the way.

Hyodar:
Thanks! Well, the idea is we need the provider data, so if it ever fails everything should fail (at least that was my initial idea). And with the way I organized it (on purpose), if anything fails while getting the data the function will fail before actually trying to write data in the DB, so we're not making the table dirty.
Still, it's not likely the logs fetching will fail, at most they'll give nothing, and that's okay. I think it's really nice to use our Goerli node if possible, does it have tracing support?

lib/indexing/utils/beaconProvider.ts
Hyodar: Ohh, actually, there was one more reason - I'm actually changing the keys, as `withdrawal_credentials` (Beacon API) -> `withdrawalCredentials` (here) so we keep the casing consistent. So, in this case, I think using the spread operator would leave too many unused keys in the object.

Hyodar:
Alright, so in [1eb3391](https://github.com/NethermindEth/restaking-dashboard/pull/10/commits/1eb33916691d33c019db8d90923c23b4c9c7c37f) I changed the flow a little bit. Now there's a function specifically for releasing the DB lock that prevents two indexing processes of the same type of happening at the same time. I already updated the Edge Functions and it seems to be working fine. Here's the current flow now:

1. Get the start and end blocks, locking the process through the DB lock (no other processes can happen at the same time)
2. Fetch the data
2.1 If fetching fails, release the lock and throw the error, failing completely
3. Update the last indexed block in the DB, but keep the lock on
4. Send the data to the respective table
5. Release the lock

The only issue I see with this is if there's a timeout while writing to the DB or there's a DB error. In both cases it would be good to keep the lock on and avoid new indexing from happening so we can properly check what happened. Another way of doing something similar would be using DB commits, but supabase JS does not support it yet, so not even sure their DB does (probably does).

If everything totally goes wrong and there's been a partial transfer to the DB, we can just get the block range from the last update with the Edge Function logs too and re-index from there.

EdgarBarrantes (approved):
LGTM, thanks for the orderly code. The functions have run smoothly for a couple of a couple of iterations.

PR #15, opened 2023-04-20
Add separate charts for beacon chain data

Hyodar:
Considering it's not exactly the same data - what we have are the current balances of the pods and how the pods were created (so we assume that's the moment the validator restaked) - I think it's better to have separate charts.

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -53,6 +53,8 @@ export default async function Home() {
     stakersStethConverted,
     groupedStakers,
     rEthRate,
+    chartDataBeaconStakesDaily,
+    chartDataBeaconStakesCumulative,
   } = await getDeposits();
 
   return (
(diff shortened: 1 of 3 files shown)


swapnilraj:
I agree now that this should be a separate chart, and it's legend/heading should be something like TVL in EigenPods

app/page.tsx
```diff
@@ -153,6 +155,36 @@ export default async function Home() {
           </div>
         </div>
 
+        <div className="charts-homepage mt-16">
+          <h3 className="text-center text-xl">
+            Cumulative Beacon Chain ETH Restaking by day
```
swapnilraj, line 160: change legend to "Cumulative Beacon Chain ETH in EigenPod by day"

swapnilraj: similar changes to the next one

EdgarBarrantes:
Added the name changes and rebased.

EdgarBarrantes (approved):
:shipit: 
There are improvements to be made to the code (not related to this PR), that can be addressed later on (I had left a note there from before).

PR #16, opened 2023-04-20
Consider rETH exchange rate in pie chart

Hyodar:
It's better to keep the number representation close to their dashboard for now, but the pie chart will only make sense if everything's ~ETH.


swapnilraj (approved):
make complete sense

PR #26, opened 2023-04-27
Add redeployment cron job for every 4 hours

EdgarBarrantes:
This is done in order to keep the data cached and loading times fast, but also keep the data updated in the site. 

This schedule runs the job at minute 0 of every fourth hour, every day of the week. So, the job will run at 12:00 AM, 4:00 AM, 8:00 AM, 12:00 PM, 4:00 PM, and 8:00 PM.


EdgarBarrantes:
Closing as we're letting supabase handle the vercel redeployment (it's an entry in the cron schema).

PR #27, opened 2023-04-27
Automated indexing through supabase edge functions

Hyodar:
Leaving it as draft while #10 is not merged.

Basically, ports the indexing functions to a Supabase project, making some adaptations because of the Deno runtime. This is exactly the reason why there's the new `generate-supabase-typechain` command, in which some `sed`s are run to make the Typechain generated typings compatible with Deno (...).
I already deployed the functions to Supabase and they all seem to be working well. In the case of the validator indexing, we were reaching the CPU time limits when trying to index everything, so instead I added some DB state and it's now incrementally indexing the validator list, 48k validators per call, going from 0 to the end and looping back once it's reached. The calls are still pretty fast, so we can just make spaced N calls on each CRON instead of an individual call.


EdgarBarrantes (approved):
Solved conflicts and reviewed code, LGTM. Probably a better README could be written in the future but added a placeholder with the docs in the meantime.

PR #38, opened 2023-06-23
Fetch TVL directly from strategy contracts

Hyodar:
This PR adds a tiny change to the FE to fetch the total locked tokens directly from the contracts, as it's not necessary to compute this from deposit data and by doing it in this way we avoid any inconsistencies between strategy shares and amounts.


yasmine9 (approved):
LGTM!

PR #40, opened 2023-06-27
Took withdrawals into account in leaderboard + share to amount conversion

yasmine9:


lib/supabaseClient.ts
```diff
@@ -1,6 +1,7 @@
 import { createClient } from "@supabase/supabase-js";
 
-const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
-const supabaseKey = process.env.NEXT_PUBLIC_SUPABASE_KEY || "";
+const supabaseUrl = "https://wsoqtgziaduigrrhnjkk.supabase.co";
+const supabaseKey =
+  "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6Indzb3F0Z3ppYWR1aWdycmhuamtrIiwicm9sZSI6ImFub24iLCJpYXQiOjE2ODE3NjE3ODIsImV4cCI6MTk5NzMzNzc4Mn0.b0EdvPtA_9EzeP1mgzfsGekk6X0DWpaWsDjJl0LoaxA";
```
Hyodar, line 5: These should probably not be here, though it's not a security issue as it's the anon one.

Hyodar (changes requested):
Great, just some small changes I suggested.

Hyodar (approved):
Made some refactoring changes and fixed a bug in the total staked amount accumulation. Also added the views to the DB. Now LGTM! Merging.

PR #50, opened 2023-08-18
added .env.example file

yasmine9:


Hyodar (approved):

Hyodar (changes requested):
Looks good, but maybe this should be merged with `example.env`.

PR #55, opened 2023-09-12
Spice ai migration

rachit2501:
- AWS Serverless function for spice ai queries
- All data replaced using spice
- Adding network switching across the application
- Dynamic rendering based on selected network
- Requires deployment before merge


api/handler.ts
```diff
@@ -0,0 +1,370 @@
...
+    SELECT
+        TO_DATE(block_timestamp) AS "date",
+        token,
+        SUM(token_amount) / POWER(10, 18) AS total_amount,
+        SUM(shares) / POWER(10, 18) AS total_shares
+    FROM eth.eigenlayer.strategy_manager_deposits
+    WHERE token IN (
+        '0xae7ab96520de3a18e5e111b5eaab095312d7fe84',
+        '0xbe9895146f7af43049ca1c1ae358b0541ea49704',
+        '0xae78736cd615f374d3085123a210448e74fc6393'
```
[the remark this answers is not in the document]
Hyodar: Similarly in other queries and in the `TokenSeries` CTEs.

api/handler.ts
```diff
@@ -0,0 +1,370 @@
...
+    GROUP BY depositor, token
+)
+    SELECT *
+    FROM ranked_deposits
+    WHERE rn <= 50;
+`);
+
+  const rEth: DailyTokenData[] = [];
+  const cbEth: DailyTokenData[] = [];
+  const stEth: DailyTokenData[] = [];
```
Hyodar, line 146: `stEthDeposits` seems better, similarly in other places.

api/handler.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  const array = response.toArray();
```
Hyodar, line 148: `deposits` seems better.

api/handler.ts
```diff
@@ -0,0 +1,370 @@
...
+    ON 
+        ac."date" = tw."date" 
+    AND 
+        (ac.token = tw.token OR (ac.token IS NULL and tw.token IS NULL))
+    ORDER BY 
+    ac."date", 
+    ac.token;
+      `);
+
+  const rEthWithdrawls: any[] = [];
```
Hyodar, line 254: `any` should also almost always be avoided.

api/handler.ts
```diff
@@ -0,0 +1,370 @@
...
+    }
+  });
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify([stEth, cbEth, rEth]),
+  };
+};
+
+export const getWithdrawals = async (
```
Hyodar, line 174: Most of the comments in `getDeposits` and `getStrategyDepositLeaderBoard` also apply here.

api/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "aws-node-http-api-typescript",
+  "description": "This template demonstrates how to make a simple HTTP API with Node.js and Typescript running on AWS Lambda and API Gateway using the Serverless Framework v3",
```
Hyodar, line 3: Boilerplate

api/spice.ts
```diff
@@ -0,0 +1,6 @@
+import { SpiceClient } from "@spiceai/spice";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+const spiceClient = new SpiceClient(process.env.SPICE || "");
```
Hyodar, line 5: Where is this env variable defined? Add an `.env.example` file with the expected keys (without the values of course).

Hyodar: Also, `SPICE_API_KEY` seems a lot better than `SPICE`.

api/yarn.lock
Hyodar, line 1: Why is there both `package-lock.json` and `yarn.lock`? This is not right, you should either be using npm or yarn and the locks can conflict.

api/handler.ts
```diff
@@ -0,0 +1,370 @@
...
+    } else if (ele.token === CBETH_ADDRESS.toLowerCase()) {
+      cbEthDeposits.push(ele);
+    } else if (ele.token === STETH_ADDRESS.toLowerCase()) {
+      stEthDeposits.push(ele);
+    }
+  });
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify([stEthDeposits, cbEthDeposits, rEthDeposits]),
```
Hyodar, line 115: Why returning an array? Similarly in all other routes.
Returning an object in this case is a lot better, so you know what's what (e.g. `{ stEthDeposits, cbEthDeposits, rEthDeposits }`).

app/utils.ts
```diff
@@ -0,0 +1,249 @@
...
+    ) / 1e18;
+  const cbEthTvl =
+    Number(
+      await cbEthStrategy.sharesToUnderlyingView(
+        await cbEthStrategy.totalShares()
+      )
+    ) / 1e18;
+
+  const depositDataPromise = axios.get<DailyTokenData[][]>(
+    `${process.env.NEXT_PUBLIC_SERVERLESS_URL}/deposits`
```
Hyodar, line 64: This env variable key should be added to `.env.example`. Also, something like `NEXT_PUBLIC_SPICE_PROXY_API_URL` seems better for me, but feel free to think more about it.

app/utils.ts
```diff
@@ -0,0 +1,249 @@
...
+const RETH_ADDRESS = "0xae78736Cd615f374D3085123A210448E74Fc6393";
+const CBETH_ADDRESS = "0xBe9895146f7AF43049ca1c1AE358B0541Ea49704";
+const STETH_STRATEGY_ADDRESS = "0x93c4b944D05dfe6df7645A86cd2206016c51564D";
+const CBETH_STRATEGY_ADDRESS = "0x54945180dB7943c0ed0FEE7EdaB2Bd24620256bc";
+const RETH_STRATEGY_ADDRESS = "0x1BeE69b7dFFfA4E2d53C2a2Df135C388AD25dCD2";
+const provider = new ethers.JsonRpcProvider("https://rpc.ankr.com/eth");
+const MAX_LEADERBOARD_SIZE = 50;
+const MAX_CHART_SIZE = 30;
+
+export async function getDashboardData() {
```
Hyodar, line 24: The requests here are not being done in client-side, but this can be done in another PR.

Hyodar: When this is done, however, remember to do all the requests async. You don't need to use `Promise.all()` for that unless you want to await all of them. You can, e.g.:
```ts
async function someAsyncFunction(): Promise<number> { return 0; }
async function someOtherAsyncFunction(): Promise<number> { return 1; }

async function main() {
  const a = someAsyncFunction(); // a is a promise
  const b = someOtherAsyncFunction(); // b is also a promise
  
  const c = await a + 1;
  const d = c + await b + await a; // no problem double awaiting a
}
```

app/utils.ts
```diff
@@ -0,0 +1,249 @@
...
+    )
+  ).data;
+
+  const stakersStEth = depositDataStakers[0] || [];
+
+  const stakersCbEth = depositDataStakers[1] || [];
+
+  const stakersREth = depositDataStakers[2] || [];
+
+  console.log(stakersStEth);
```
Hyodar, line 176: Debugging log.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 3001",
```
Hyodar, line 6: ?

lib/spiceAi.ts
Hyodar, line 1: Doesnt't make sense to have this here anymore.

Hyodar (changes requested):

api/handler.ts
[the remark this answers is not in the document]
Hyodar: Where are the cumulative queries?

app/utils.ts
```diff
@@ -0,0 +1,249 @@
...
+    ele.map((value, index, arr) => {
+      const cumulativeAmount =
+        index == 0
+          ? value.total_amount
+          : value.total_amount! + arr[index - 1].total_amount!;
+      // @ts-ignore
+      value.total_amount = cumulativeAmount;
+      return value;
+    })
+  );
```
Hyodar, line 106: We'd previously talked about this. This should not be done client-side unless there's a really good reason for that, and the way you did is actually a server-side build, so it doesn't make sense at all. Also, from my tests here, adding the cumulative values as columns the deposits and withdrawals queries themselves (it's just a sum over with partition) didn't change the query time at all.

Hyodar (changes requested):

package.json
rachit2501: This was required since the api server was using port 3000

api/handler.ts
Hyodar, line 1: I think this query will not work for the cumulative values, please check the query in our discord conversation.

api/handler.ts
```diff
@@ -0,0 +1,428 @@
...
+  `);
+
+  const rEthDeposits: DailyTokenData[] = [];
+  const cbEthDeposits: DailyTokenData[] = [];
+  const stEthDeposits: DailyTokenData[] = [];
+  const beaconChainDeposits: DailyTokenData[] = [];
+
+  const array = response.toArray();
+
+  // @ts-ignore
```
Hyodar, line 122: Would be better to quickly type this, it's not a hard type.

api/handler.ts
```diff
@@ -0,0 +1,428 @@
...
+      statusCode: 200,
+      body: JSON.stringify(result.toArray()),
+    };
+  } catch (err: any) {
+    console.log(err);
+    throw err;
+  }
+}
+
+export async function dailyBeaconChainETHDeposit() {
```
Hyodar, line 377: This should not be needed with the functions above. We could also make the validator queries separately if there's some gain in perf, but let's keep it simple on this PR.

api/package.json
Hyodar, line 1: Typos: Strip, stripe. Actually, there are these typos all over the place.

example.env
Hyodar, line 1: Not sure I got these changes. Supabase url and key shouldn't be public (neither does the api url for now, it's only needed for exposing it to the client). Also, stripe typo.

lib/utils.ts
```diff
@@ -1,14 +1,36 @@
-
 export interface DailyTokenData {
   date: string | null;
   total_amount: number | null;
-};
+  total_shares: number | null;
```
Hyodar, line 4: Is everything nullable? Maybe those `| null` can be dropped.

lib/utils.ts
```diff
@@ -1,14 +1,36 @@
...
+}
+
+export interface Deposits {
+  stEthDeposits: DailyTokenData[];
+  cbEthDeposits: DailyTokenData[];
+  rEthDeposits: DailyTokenData[];
+  beaconChainDeposits: DailyTokenData[];
+}
+
+export interface Withdraws {
```
Hyodar, line 27: Withdraws, withdrawls -> withdrawals. Multiple cases of this in other files I guess.

api/handler.ts
```diff
@@ -0,0 +1,428 @@
...
+      case RETH_ADDRESS:
+        rEthDeposits.push(ele);
+        break;
+      case CBETH_ADDRESS:
+        cbEthDeposits.push(ele);
+        break;
+      case STETH_ADDRESS:
+        stEthDeposits.push(ele);
+        break;
+      default:
```
Hyodar, line 134: This shouldn't be a default IMO, this is a case null. Default would mean any other token address maps to a beacon chain deposit.

api/handler.ts
```diff
@@ -0,0 +1,428 @@
...
+      rEthWithdrawls,
+      beaconChainWithdrawls,
+    }),
+  };
+};
+
+export async function totalStakedBeaconChainEth() {
+  try {
+    const result =
+      await SpiceClient.query(`SELECT SUM(effective_balance)/POW(10,9) as final_balance
```
Hyodar, line 338: `SpiceClient` should probably be `spiceClient`. Multiple cases of this.

Hyodar (changes requested):
Not an exhaustive review, just what I could check today. Some more suggestions, lmk wyt.

api/handler.ts
rachit2501: the query only tackles 4 cases right? so default will always be null based on that.

Hyodar: Yes, and no. You 'can' think it's like that, but there's no notable benefit from doing this and it can lead to issues - e.g. a new token would be put in the beacon data basket.
A default is usually implemented as a fallback, not as an else, and in this case even an else would not be nice.

Hyodar (approved):

PR #58, opened 2023-11-09
Update the workflow permissions for GITHUB_TOKEN

adriantpaez:


Hyodar (approved):

adriantpaez:
I will close this PR because it is fixing a deployment configuration that should be removed in favor of the one added in #59.

PR #61, opened 2023-11-15
Timeline selection

tdenisenko:
- Timeline selection and grouping for charts
- Some refactoring


api/package-lock.json
Hyodar, line 1: Hey, please use `yarn` on this - having both `yarn.lock` and `package-lock.json` can lead to some dep issues.

api/serverless.yml
```diff
@@ -16,7 +16,7 @@ functions:
   getDeposits:
     handler: handler.getDeposits
     runtime: nodejs18.x
-    timeout: 30
+    timeout: 40
```
Hyodar, line 19: Usually API Gateway's integration has a hard limit on 30 seconds, so I don't think it makes sense that we increase it to 40s, which should work only locally. Has anything changed and API Gateway now supports > 30s?

app/components/hooks/useDeposits.ts
```diff
@@ -1,20 +1,20 @@
...
 
-export function getDepositsQueryKey(network: SupportedNetwork): any[] {
-  return ["deposits", network];
+function getDepositsQueryKey(network: SupportedNetwork, timeline: Timeline): any[] {
+  return ["deposits", network, timeline];
 }
 
-export async function queryDeposits(network: SupportedNetwork, _: boolean = false): Promise<ApiDepositsResponse> {
-  return await getDeposits(network);
+async function queryDeposits(network: SupportedNetwork, timeline: Timeline, _: boolean = false): Promise<ApiDepositsResponse> {
```
Hyodar, line 10: Good! Maybe we can think about using cached entries from bigger ranges in the future? I don't think this needs to be in this PR though.

app/components/hooks/useDepositsGrouping.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  data.timestamps.forEach((timestamp, i) => {
+    const entryDate = moment(timestamp);
+
+    let time
+
+    if (timeRange === "weekly") {
+      // const weekStart = entryDate.clone().startOf('week');
+      // const weekEnd = entryDate.clone().endOf('week');
+      // time = `${weekStart.format('YYYY-MM-DD')} to ${weekEnd.format('YYYY-MM-DD')}`;
```
Hyodar, line 37: Maybe should be removed?

app/components/hooks/useWithdrawalsGrouping.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  data.timestamps.forEach((timestamp, i) => {
+    const entryDate = moment(timestamp);
+
+    let time
+
+    if (timeRange === "weekly") {
+      // const weekStart = entryDate.clone().startOf('week');
+      // const weekEnd = entryDate.clone().endOf('week');
+      // time = `${weekStart.format('YYYY-MM-DD')} to ${weekEnd.format('YYYY-MM-DD')}`;
```
Hyodar, line 36: Should probably also be removed

app/utils/types.ts
```diff
@@ -9,12 +9,19 @@ export type SupportedNetwork = (typeof supportedNetworks)[number];
 
 export type TokenRecord<T> = Record<SupportedToken | "beacon", T>;
 
+export const timeline = ["1w", "1m", "1y", "full"] as const;
```
Hyodar, line 12: I think `supportedTimelines` or `timelines` would be a better name. I see functions that receive `timeline` that would quite easily clash with this, so it gets confusing.

next.config.js
```diff
@@ -1,4 +1,16 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
 
-module.exports = nextConfig
+const nextConfig = {
+    // reactStrictMode: true,
```
Hyodar, line 4: Should probably be removed

next.config.js
```diff
@@ -1,4 +1,16 @@
...
+};
+
+module.exports = {
+    ...nextConfig,
+    webpack: (config, { isServer }) => {
+        if (!isServer) {
+            config.externals.push({ bufferutil: "bufferutil", "utf-8-validate": "utf-8-validate", "supports-color": "supports-color" });
```
Hyodar, line 11: Can you explain this?

package-lock.json
Hyodar, line 1: Ref. comment about `yarn`

package.json
```diff
@@ -40,7 +41,10 @@
     "@typechain/ethers-v6": "^0.3.0",
     "@types/lodash": "^4.14.194",
     "@types/react-modal": "^3.16.0",
+    "bufferutil": "^4.0.7",
     "encoding": "^0.1.13",
-    "typechain": "^8.1.1"
+    "supports-color": "^8.1.1",
+    "typechain": "^8.1.1",
+    "utf-8-validate": "^5.0.10"
```
Hyodar, line 48: This is probably related to the webpack changes. What is the idea?

Hyodar (changes requested):
Hey, good work! Some small points, though nothing too major.

api/package-lock.json
tdenisenko: Agreed, will remove package-lock.json

api/serverless.yml
tdenisenko: This was required while testing since I got too many timeout errors. It seems like queries got faster now, will switch back to 30s

app/components/hooks/useDeposits.ts
tdenisenko: This was the main requirement for this branch, we want to be able to query different timelines

app/components/hooks/useDepositsGrouping.ts
tdenisenko: Will remove

app/components/hooks/useWithdrawalsGrouping.ts
tdenisenko: Will remove

app/utils/types.ts
tdenisenko: That makes sense

next.config.js
tdenisenko: Will remove

next.config.js
tdenisenko: This fixed an error message thrown out regarding the bufferutil but I can probably revert this completely to its initial state

package-lock.json
tdenisenko: Will remove

package.json
tdenisenko: Same explanation regarding `bufferutil` I will try to revert this

tdenisenko:
Requested changes made, please review

Hyodar (approved):
Added caching based on bigger time ranges, did some small changes and merged `dev` into here. Feel free to double-check and then merge into `dev`.

Do note the RPC URL env keys renaming on your `.env`.

PR #62, opened 2023-11-22
Add tags to serverless resources

adriantpaez:


api/package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "@spiceai/spice": "^0.5.1",
     "dotenv": "^16.3.1",
+    "serverless-plugin-resource-tagging": "^1.2.0",
```
Hyodar, line 9: Should this maybe be in `devDependencies`?

Hyodar (approved):
Hey! Maybe there is a dep that needs to be moved to dev, but that's it. I'll leave it to you!

api/package.json
adriantpaez: I followed the instructions in the serverless documentation to install this plugin, but I noticed other serverless plugins and dependencies are in devDependencies. I will move it and run the deployment again to check if all is ok

Issue #88, opened 2024-05-20
fix: ui issues and enhancements


kushagrasarathe:
These UI issues and enhancements needs to be implemented: 

- Sidebar also scrolls-down along with the page content, hiding the sidebar title. Ideally sidebar should be fixed wrt to page content, keeping all links and title visible on screen without scrolling.

<img width="613" alt="Screenshot 2024-05-20 at 10 13 04 AM" src="https://github.com/NethermindEth/restaking-dashboard/assets/76868364/4a84e713-9bde-4449-89ec-886747f52838">

<hr/>

- Padding between tabs sections on smaller screens
<img width="244" alt="Screenshot 2024-05-20 at 11 08 35 AM" src="https://github.com/NethermindEth/restaking-dashboard/assets/76868364/b710eac3-903c-4cae-adf5-a62cef9b99d1">

<hr/>

- Fix mobile responsiveness for overflowing content 

<hr/>

- Close ham menu on smaller screens on navigating to a different path

Abhii5496:
hey , can i work on this ?

kushagrasarathe:
> hey , can i work on this ?

Hey @Abhii5496 I appreciate your willingness to contribute. I am currently working on this issue myself and it's in progress.

Thanks, 
Kushagra

David-Mastalski:
So, instead of one button.js component that performs various events, you would like to create several buttons where each button will be responsible for something different? I can take care of it if you want

PR #92, opened 2024-06-03
fix: ui issues 

kushagrasarathe:
closes: #88

- fix sidebar scrolling when layout content exceeds view-port
- fix sidebar modal to close on navigating to different path
- fix layout overflow
- fix mobile padding on avs page


.gitignore
```diff
@@ -29,3 +29,5 @@ dist/
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+
+.env
```
rubo, line 33: Why is this?

kushagrasarathe: We wanna have the `env` public? Normally we keep env private to avoid pushing sensitive api keys. Lemme know I can remove it back.

src/shared/Layout.jsx
```diff
@@ -15,7 +15,7 @@ export default function Layout() {
       )}
       <div className="basis-0 flex-1">
         {!showSidebar && <Header />}
-        <main className="p-4 lg:p-6">
+        <main className="p-4 lg:p-6 max-w-[100vw]">
```
vincenthongzy, line 18: might be a nitpick, but on desktop browsers there is also an horizontal scrollbar for the contents (probably due to the Sidenav taking up precious screen real estate). But we can discuss with Carlos on Wed

src/shared/Layout.jsx
```diff
@@ -7,14 +7,14 @@ export default function Layout() {
   const showSidebar = useTailwindBreakpoint('lg');
 
   return (
-    <div className="bg-background min-h-screen flex flex-col text-foreground">
-      <div className="basis-0 flex flex-1">
+    <div className="bg-background min-h-screen flex flex-col text-foreground relative">
+      <div className="basis-0 flex flex-1 ">
         {showSidebar && (
-          <div className="flex w-64">
+          <div className="flex w-64 h-screen fixed z-50 lg:w-64">
```
rubo, line 13: Why's `z-50`?

rubo (approved):

src/shared/Layout.jsx
kushagrasarathe: had to add it to keep the sidebar on top since have added a fixed class, otherwise the page content might overlap with sidebar. 

like this, this is just for demo, otherwise the padding would also prevent from overlapping 
Screenshot 2024-06-26 at 7 17 16 PM

PR #93, opened 2024-06-11
Add AVS page

kushagrasarathe:

PR #98, opened 2024-06-24
feat: avs page graphs with api integration

kushagrasarathe:
- avs tvl over time api integration
- tvl overtime graph component updated to render api data
- operators over time api integration and graph


src/@services/AVSService.js
```diff
@@ -37,4 +37,15 @@ export default class AVSService {
     // TODO: Handle error
     return await response.json();
   }
+
+  async getAvsTvlOvertime(address) {
```
vincenthongzy, line 41: there is this comment [here](https://github.com/NethermindEth/restaking-dashboard/pull/93#discussion_r1655157835) about following naming conventions, but it would be ridiculous to name this `getAVSTVLOvertime`. Ruben suggested to use `getAVSTotalValue` and drop the "Overtime" 
```suggestion
  async getAVSTotalValue(address) {
```

src/avs/AVSDetails.jsx
```diff
@@ -273,31 +273,8 @@ export default function AVSDetails() {
...
-                <GraphTimelineSelector
-                  timelineTab={state.timelineTab}
-                  onTimelineChange={handleTimelineChange}
-                />
-              </CardHeader>
-              <CardBody>
-                <TVLOverTime />
-              </CardBody>
-            </Card>
+            <AVSTVLOverTime avsAddress={avsAddress} />
```
vincenthongzy, line 276:
```suggestion
            <AVSTotalValueOvertime avsAddress={avsAddress} />
```

src/utils.js
```diff
@@ -46,3 +46,36 @@ export const assetFormatter = new Intl.NumberFormat('en-US', {
...
+    return (value / 1e9).toFixed(1) + 'B';
+  }
+  if (value >= 1e6) {
+    return (value / 1e6).toFixed(1) + 'M';
+  }
+  if (value >= 1e3) {
+    return (value / 1e3).toFixed(1) + 'K';
+  }
+  return value.toFixed(1);
+};
```
vincenthongzy, line 81: don't we already have a `formatNumber` in the same file?

kushagrasarathe: the `formatNumber` is being used on multiple places in the app, and using the same function for graphs is breaking the ui sometimes, and `formatNumberToCompactString` I created specifically for using in graphs to avoid breaking the numbers in any other places. 

Screenshot 2024-06-29 at 3 12 35 PM

vincenthongzy:
```
  if (compact) {
    if (value >= 1_000_000) {
      return `${(value / 1_000_000).toFixed(1)}m`;
    }
    if (value >= 1_000) {
      return `${(value / 1_000).toFixed(1)}k`;
    }
  }
  return numberFormatter.format(value);
```

which branch of formatNumber is breaking the scale? seems like it should have return  `XXXk` instead of the whole number?

vincenthongzy: might be in our best interest to have all the graphs scales behave and look the same way. Because with your new function, some are in upper-case and some are in lower-case.

kushagrasarathe: removing this function, my bad for not observing this before, fixed it by simply using the `formatNumber` function with a `true` compact param for axis

src/avs/AVSTotalValueOvertime.jsx
```diff
@@ -201,9 +201,9 @@ const AVSTVLOverTime = ({ avsAddress }) => {
                   numTicks={getNumberOfTicks(width, 'x')}
                 />
                 <AxisLeft
-                  scale={yScale}
-                  top={margin.top}
-                  left={margin.left}
+                  scale={tvlScale}
+                  top={margin.top + 15}
+                  left={margin.left - 15}
```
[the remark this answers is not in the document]
kushagrasarathe: fixed this one in latest commit

vincenthongzy (approved):

PR #99, opened 2024-06-27
feature: add operators over time graph

kushagrasarathe:
- adds operators over time graph and api integration 

Screenshot 2024-06-29 at 2 50 50 PM

----- 

Note: @rubo @vincenthongzy temporarily removed latest operators and inactive operators section from UI to avoid rendering incorrect or mock data, will add it once we have sorting ready on backend.

The AVS Operators section now takes whole width
Screenshot 2024-07-02 at 1 12 13 AM


vincenthongzy:
Is this graph showing the daily number of registered operators:
```
day 1: 1 new operator
day 2: 2 new operator
day 3: 3 new operator (does not show deregistration)
```
 or is this a graph showing the accumulation of operators over time?
```
day 1: 1 total
day 2 : 3 total
day 3: 5 total ( 1 de-registered from day2)
```

src/avs/AVSDetails.jsx
```diff
@@ -291,21 +291,7 @@ export default function AVSDetails() {
...
-                  timelineTab={state.timelineTab}
-                  onTimelineChange={handleTimelineChange}
-                />
-              </CardHeader>
-
-              <CardBody>
-                <OperatorsOverTime />
-              </CardBody>
-            </Card>
+            <OperatorsOverTime avsAddress={avsAddress} />
```
vincenthongzy, line 294: Is this graph showing the daily number of registered operators:
```
day 1: 1 new operator
day 2: 2 new operator
day 3: 3 new operator (does not show deregistration)
```
 or is this a graph showing the accumulation of operators over time?
```
day 1: 1 total
day 2 : 3 total
day 3: 5 total ( 1 de-registered from day2)
```

kushagrasarathe: @vincenthongzy  I'm not formatting the data on frontend, it's being represented the way it's received from backend. So I believe it's the later one (accumulation of operators over time).

Did I get your question right though? 👀

src/avs/OperatorsOverTime.jsx
```diff
@@ -1,99 +1,278 @@
...
+    if (!sortedData) return null;
+    const maxValue = Math.max(...sortedData.map(d => d.operators));
+    const minValue = Math.min(...sortedData.map(d => d.operators));
+
+    // create artificial range if all values same
+    const yDomain =
+      maxValue === minValue
+        ? [maxValue * 0.9, maxValue * 1.1]
+        : [minValue, maxValue + (maxValue - minValue) * 0.1];
+
```
vincenthongzy, line 120: Hm, how can all the values be the same? When the AVS is new and we have less than a few days worth of data?  

If so, could you update the comment and leave this context in ?

vincenthongzy (commented):
The design doesn't seem to show verbose dates with the month, just the day. Is this change communicated with Carlos?

image

src/avs/OperatorsOverTime.jsx
```diff
@@ -1,99 +1,278 @@
...
-            textAnchor: 'middle'
-          })}
+    <Card
+      radius="md"
+      className="bg-content1 border border-outline p-4 relative"
+    >
+      <CardHeader className="flex items-end flex-wrap justify-between gap-3">
+        {' '}
+        <div className="font-light text-lg text-foreground-1">
+          Operators over time
```
vincenthongzy, line 166: as discussed, we need to clarify with product what this graph is trying to represent here. Daily registrations over time or total operator count over time. Please update the title and leave a comment here when that's sorted out @kushagrasarathe

src/avs/OperatorsOverTime.jsx
```diff
@@ -1,99 +1,278 @@
...
+        return data.slice(-7);
+      case '30days':
+        return data.slice(-30);
+      default:
+        return data;
+    }
+  };
+
+  const sortedData = useMemo(() => {
+    if (!graphData) return null;
```
vincenthongzy, line 93: hm when would this be null?

kushagrasarathe: if the data is not fetched from the api, the default value is null, so this check was added to prevent breaking the page

src/avs/OperatorsOverTime.jsx
kushagrasarathe: this needs to be `total operator count over time`, I'm working on fixing this 🫡

src/avs/OperatorsOverTime.jsx
```diff
@@ -1,99 +1,278 @@
...
+  const handleTimelineChange = useCallback(
+    tab => {
+      dispatch({ timelineTab: tab });
+    },
+    [dispatch]
+  );
+
+  const handleMouseMove = useCallback(
+    event => {
+      if (!sortedData || !xScale || !yScale) return;
```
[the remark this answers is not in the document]
kushagrasarathe: @vincenthongzy we can do that, but the reason I chose x and y scales as names was to easily distinguish them based on the axis they are for.

kushagrasarathe:
> The design doesn't seem to show verbose dates with the month, just the day. Is this change communicated with Carlos?
> 
> image

@vincenthongzy I improvised here, cuz user can have a better context if we show the verbose dates, instead of just day, otherwise things can be a little confusing from a users pov if I don't even know what the metrics is being represented based on, thoughts?

kushagrasarathe:
oops closed the pr by mistake!

src/avs/OperatorsOverTime.jsx
```diff
@@ -1,99 +1,278 @@
...
+  const sortData = (data, tab) => {
+    switch (tab) {
+      case '7days':
+        return data.slice(-7);
+      case '30days':
+        return data.slice(-30);
+      default:
+        return data;
+    }
+  };
```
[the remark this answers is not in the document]
kushagrasarathe: I think there's no need to wrap it in a `useCallback` since it doesn't depend on any state/prop or have a side-effect, its more of a utility function.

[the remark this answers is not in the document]
kushagrasarathe: `sortTimelineData` would be a better name?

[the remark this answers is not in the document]
kushagrasarathe: > I think there's no need to wrap it in a `useCallback` since it doesn't depend on any state/prop or have a side-effect, its more of a utility function.

my bad, the function depends on timeline tab state, have wrapped it inside a `useCallback`

src/avs/OperatorsOverTime.jsx
kushagrasarathe: @vincenthongzy  I added this check to create a artificial range for y-axis cuz there's a case when the operators count is constant throughout different dates, which makes the y-axis render only single tick value, which looks kind of weird, so for that case I'm creating a range. See below screenshot of how it looks without this range

Screenshot 2024-06-29 at 2 42 24 PM

src/avs/OperatorsOverTime.jsx
kushagrasarathe: update: I've updated the backend pr to use `avs_snapshots` collection, lemme know if it needs any changes and if the approach is correct. Thanks

vincenthongzy (approved):

src/avs/OperatorsOverTime.jsx
```diff
@@ -260,14 +263,16 @@ const OperatorsOverTime = ({ avsAddress }) => {
           {tooltipOpen && tooltipData && (
             <TooltipInPortal
               key={Math.random()}
-              top={tooltipTop}
-              left={tooltipLeft}
+              top={tooltipTop + 60}
+              left={tooltipLeft - 250}
```
vincenthongzy, line 267: why the fixed pixel values here? :o

src/avs/OperatorsOverTime.jsx
vincenthongzy, line 1: image

did you rename the file? seems like its broken for me locally

vincenthongzy (changes requested):

src/avs/OperatorsOverTime.jsx
vincenthongzy: image

seems like its causing the tooltip to be out of place

vincenthongzy: image

same problem with @Zishan-7 PR, the tooltips are not showing correctly

vincenthongzy: another example of it behaving weirdly 

https://github.com/NethermindEth/restaking-dashboard/pull/100#discussion_r1660746982

src/avs/OperatorsOverTime.jsx
vincenthongzy: image

dates are broken on smaller screens + the y axis should be whole numbers and not have `.0`

kushagrasarathe: @vincenthongzy  working on replacing all the graphs to follow similar structure and fixing tooltip with Zishan's approach. other than that everything else looks good on this PR to you?

src/avs/OperatorsOverTime.jsx
kushagrasarathe: have pushed a fix, tooltip is working fine now

src/avs/OperatorsOverTime.jsx
```diff
@@ -132,151 +50,31 @@ const OperatorsOvertime = ({ avsAddress }) => {
...
-        tooltipData: d,
-        tooltipLeft: x,
-        tooltipTop: operatorsScale(d.operators)
-      });
-    },
-    [showTooltip, sortedData, dateScale, operatorsScale, bisectTimestamp]
-  );
+  useEffect(() => {
+    fetchOperatorsOvertime();
+  }, [avsService, dispatch, avsAddress]);
```
vincenthongzy, line 55: We should move the function definition in here. Otherwise this dependency array is wrong, it has to be `[fetchOperatorOvertime]`, and you need to wrap `fetchOperatorOvertime` in `useCallback`. Better move it here as a IIFE

src/avs/OperatorsOverTimeChart.jsx
```diff
@@ -0,0 +1,214 @@
...
+      const { x, y } = localPoint(ev) || { x: 0, y: 0 };
+      const date = dateScale.invert(x - margin.left);
+
+      showTooltip({
+        tooltipData: getOperatorsByDate(date),
+        tooltipLeft: x,
+        tooltipTop: y
+      });
+    },
+    [localPoint, dateScale, margin, showTooltip, getOperatorsByDate]
```
vincenthongzy, line 93: `[localPoint, dateScale, margin, showTooltip, getOperatorsByDate]`

getOperatorsByDate  is not wrapped in `useCallback`, so its doing nothing in the dependency array here

src/avs/OperatorsOverTime.jsx
```diff
@@ -1,100 +1,83 @@
...
+      <CardBody className="w-full h-[400px]">
+        <ParentSize>
+          {({ width, height }) => (
+            <OperatorsOvertimeChart
+              data={filteredData}
+              width={width}
+              height={height}
+            />
+          )}
+        </ParentSize>
```
vincenthongzy, line 77: Hey can you double check if its responsive? I tried it on my end but it just gives a horizontal overflow instead. There is no resizing of the charts happening

image

src/avs/Operators.jsx
```diff
@@ -150,37 +101,45 @@ export default function Operators({ avsAddress, totalTVL }) {
...
-                  {(
-                    (parseFloat(operator.strategiesTotal) / totalTVL) *
-                    100
-                  ).toFixed(2)}
-                  %
-                </span>
-                <span className="basis-1/3 text-end">
-                  <div>
-                    {formatNumber(operator.strategiesTotal, compact)} ETH
+          {filteredOperators.length ? (
```
vincenthongzy, line 107: here is a issue here, if you go to the next page, you will see that the operator's logo image does not change immediately. Either we fix this or add a skeleton.

vincenthongzy (changes requested):

src/avs/Operators.jsx
kushagrasarathe: @vincenthongzy  does this look good?

https://github.com/NethermindEth/restaking-dashboard/assets/76868364/cff8eff5-fed7-4ae7-96e7-07333ce0b597

vincenthongzy: nice!

src/avs/OperatorsOverTimeChart.jsx
```diff
@@ -90,7 +90,7 @@ const OperatorsOvertimeChart = ({ data, width, height }) => {
         tooltipTop: y
       });
     },
-    [localPoint, dateScale, margin, showTooltip, getOperatorsByDate]
```
vincenthongzy: you need to wrap `getOperatorsByDate` in a useCallback. Not remove it, or the useCallback dependency array is wrong. If its not used anywhere, can possibly just move it in this chunk

kushagrasarathe: moved it inside the function, was only being used in one place

vincenthongzy (approved):

src/avs/OperatorsOvertimeChart.jsx
```diff
@@ -169,8 +169,8 @@ const OperatorsOvertimeChart = ({ data, width, height }) => {
             {tooltipData && (
               <g>
                 <Circle
-                  cx={dateScale(new Date(tooltipData.timestamp))}
-                  cy={operatorsScale(tooltipData.operators)}
+                  cx={tooltipLeft - margin.left}
```
vincenthongzy, line 172: since you minus off margin.left here, maybe you don't need to add it in in line 88?

vincenthongzy (approved):
good stuff

src/avs/OperatorsOvertimeChart.jsx
kushagrasarathe: @vincenthongzy  actually it is important to add, otherwise I'm not able to move the cursor to last data-point and see it's tooltip, see below screenshots, 1st one is with addition of `margin.left` and second one is wh

Screenshot 2024-07-03 at 12 20 45 AM

Screenshot 2024-07-03 at 12 21 13 AM

vincenthongzy: oh ok my bad, this is for the circle svg, not the tooltip. All's good, thanks!

PR #100, opened 2024-06-28
Integrate Operator TVL over time API and update restakers trend chart (Operators Page)

Zishan-7:
- Integrated operators TVL API
- Updated Restakers trend chart and integrated API


src/operators/RestakersTrend.jsx
```diff
@@ -0,0 +1,82 @@
...
+  }, [state.restakerTrend, state.timelineTab]);
+
+  const handleTimelineChange = tab => {
+    dispatch({ timelineTab: tab });
+  };
+
+  const getRestakerTrend = async () => {
+    const restakerData = await operatorService.getRestakerTrend(opAddress);
+    dispatch({ restakerTrend: restakerData });
+  };
```
vincenthongzy, line 40: since you only used this once, it can be inside useEffect as an IIFE.

src/operators/RestakersTrend.jsx
```diff
@@ -0,0 +1,82 @@
...
+    <Card radius="md" className="bg-content1 w-full border border-outline p-4">
+      <CardHeader className="flex flex-wrap justify-between gap-3">
+        <div className="space-y-2 block">
+          <span className="text-foreground-2">Restaker Trend</span>
+          <div className="font-light">
+            <div className="text-base text-foreground-1">
+              <span>
+                Total Restakers:{' '}
+                {state.restakerTrend.length != 0 &&
+                  state.restakerTrend[state.restakerTrend.length - 1].restakers}
```
vincenthongzy, line 56: maybe return a N/A if its empty for some reason

src/operators/RestakersTrendChart.jsx
```diff
@@ -0,0 +1,204 @@
...
+  const handleTooltip = useCallback(event => {
+    const { x, y } = localPoint(event) || { x: 0, y: 0 };
+    const date = dateScale.invert(x - margin.left);
+
+    showTooltip({
+      tooltipData: getRestakersByDate(date),
+      tooltipLeft: x,
+      tooltipTop: y
+    });
+  });
```
vincenthongzy, line 82: you are missing the dependency array for useCallback here

src/operators/RestakersTrendChart.jsx
```diff
@@ -0,0 +1,204 @@
...
+        >
+          <div className="text-sm">
+            Date: {formatDateToVerboseString(new Date(tooltipData.timestamp))}
+          </div>
+          <div className="text-base">Restakers: {tooltipData.restakers}</div>
+        </TooltipWithBounds>
+      ) : null}
+    </div>
+  );
+};
```
vincenthongzy, line 202: image

tooltips shouldn't go out of bounds, it should reverse its direction

src/operators/RestakersTrendChart.jsx
```diff
@@ -0,0 +1,204 @@
...
+            toOpacity={0.1}
+          />
+
+          <AxisLeft
+            tickTextFill={'#7A86A5'}
+            scale={restakersScale}
+            tickLabelProps={() => ({
+              fill: '#7A86A5',
+              fontSize: 11,
+              textAnchor: 'end'
```
vincenthongzy, line 126: `nit`: we really should be using tailwind variables like font-xs and maybe define #7A86A5 in the tailwind config file.

src/operators/OperatorDetails.jsx
```diff
@@ -133,28 +111,7 @@ const OperatorDetails = () => {
...
-              <GraphTimelineSelector
-                timelineTab={state.timelineTab}
-                onTimelineChange={handleTimelineChange}
-              />
-            </CardHeader>
-            <CardBody>
-              <RestakerTrend width={670} height={400} />
-            </CardBody>
-          </Card>
+          <RestakersTrend opAddress={address} />
```
vincenthongzy, line 114: image

doesn't really look nice on mobile, let's grab the low hanging fruit and give it a row based layout on small screens

src/operators/RestakersTrendChart.jsx
vincenthongzy: curious, why TooltipWithBounds here but TooltipWithPortal in the `OperatorTVLOverTime` component?

src/operators/OperatorTVLOverTime.jsx
```diff
@@ -0,0 +1,295 @@
...
+  }, [filteredData, height, margin]);
+
+  const handleTimelineChange = useCallback(
+    tab => {
+      dispatch({ timelineTab: tab });
+    },
+    [dispatch]
+  );
+
+  const handleMouseMove = React.useCallback(
```
vincenthongzy, line 114: import at the top
```suggestion
  const handleMouseMove = useCallback(
```

src/operators/OperatorTVLOverTime.jsx
```diff
@@ -0,0 +1,295 @@
...
+      const d0 = filteredData[index - 1];
+      const d1 = filteredData[index];
+      let d = d0;
+      if (d1 && d1.timestamp) {
+        d =
+          x0.valueOf() - new Date(d0.timestamp).valueOf() >
+          new Date(d1.timestamp).valueOf() - x0.valueOf()
+            ? d1
+            : d0;
+      }
```
vincenthongzy, line 130: i don't think the logic here is working well, its kinda funky, i move it a little bit and its jumping between 3 points

https://github.com/NethermindEth/restaking-dashboard/assets/107991825/4bc6e228-5918-4535-a82a-6ab4dc0fe262

src/operators/OperatorTVLOverTime.jsx
```diff
@@ -0,0 +1,295 @@
...
+      dispatch({
+        tvlOvertimeData
+      });
+    } catch (error) {
+      // TODO: handle error
+    }
+  };
+
+  useEffect(() => {
+    fetchTVLOverTime();
```
vincenthongzy, line 153: are we calling it `Overtime` or `OverTime` ?

src/operators/OperatorTVLOverTime.jsx
```diff
@@ -0,0 +1,295 @@
...
+        <div
+          ref={containerRef}
+          className={cn('w-full', `h-[${height}px] max-h-[400px] relative`)}
+        >
+          <svg
+            width="100%"
+            height={height}
+            viewBox={`0 0 ${width} ${height}`}
+            onMouseMove={handleMouseMove}
+            onMouseLeave={() => hideTooltip()}
```
vincenthongzy, line 183:
```suggestion
            onMouseLeave={hideTooltip}
```

vincenthongzy (changes requested):

vincenthongzy:
priority to address the UI/UX , can leave the `nit` for later

src/operators/RestakersTrendChart.jsx
vincenthongzy, line 1: image

highly unlikely to have constant value over 7 days, but maybe can take some inspiration from https://github.com/NethermindEth/restaking-dashboard/pull/99#discussion_r1658112437 to create artificial ranges to make it look more natural

src/operators/OperatorTVLOverTime.jsx
vincenthongzy: we are calling it `Overtime` in Kushagra's PR

src/operators/RestakersTrendChart.jsx
Zishan-7: Actually I used it to counter the same bug you mentioned. `TooltipWithBounds` is aware of its parent's boundaries and it flips if it is going out of bounds.

src/operators/OperatorTVLOverTime.jsx
Zishan-7: This is a different chart component than that of Restakers Trend chart. We are aware of this tooltip bug and @kushagrasarathe is working to fix this. If the Restaker Trend chart looks okay to you then we can use that component instead of this one.

src/operators/RestakersTrendChart.jsx
vincenthongzy: It doesn't seem like its working 😅

Zishan-7: Fixed it, working now

Screenshot 2024-07-01 at 3 53 10 PM

src/operators/OperatorTVLOverTime.jsx
vincenthongzy: @Zishan-7 can confirm the Restakers Trend chart is behaving as expected (apart from the out of bounds tooltip), but the actual tooltip behavior is sane.

Zishan-7: The tooltip bug is fixed. I will use that chart here as well

src/operators/OperatorTVLOverTime.jsx
Zishan-7: Converted it to IIFE

src/operators/RestakersTrendChart.jsx
```diff
@@ -86,7 +85,7 @@ const RestakersTrendChart = ({ data, width, height }) => {
 
   return (
     <div>
-      <svg width={width} height={height}>
+      <svg width={width} height={height} className="overflow-visible">
```
vincenthongzy, line 88: hm what's with the overflow-visible changes here? do you have any screenshots?

Zishan-7: The y-axis numbers get cut off without `overflow-visible`


Screenshot 2024-07-02 at 1 15 42 PM


Screenshot 2024-07-02 at 1 15 55 PM

vincenthongzy (approved):

PR #102, opened 2024-07-01
feat: avs list pagination

kushagrasarathe:
- adds pagination to `AVSList` component
- also fixes the issue with breaking pages when metadata is not available for a AVS by not rendering them in the list

@vincenthongzy the fix for AVSs with no metadata is temporary on frontend to prevent the page from breaking, we can remove this once we fix it on the backend

### DIFF
--- a/src/@services/AVSService.js
+++ b/src/@services/AVSService.js
@@ -1,8 +1,13 @@
 import { apiGet } from './apiCall';
 
 export default class AVSService {
-  async getAll() {
-    const response = await apiGet('/avs');
+  async getAll(pageNumber) {
+    const pageIndex = Math.max(0, pageNumber - 1);
+    const response = await apiGet('/avs', {
+      query: {
+        'page-index': pageIndex
+      }
+    });
 
     if (response.ok) {
       return await response.json();
--- a/src/avs/AVSList.jsx
+++ b/src/avs/AVSList.jsx
@@ -1,14 +1,16 @@
 import { Skeleton } from '@nextui-org/react';
 import { useEffect } from 'react';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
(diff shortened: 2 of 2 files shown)


src/@services/AVSService.js
```diff
@@ -1,8 +1,13 @@
 import { apiGet } from './apiCall';
 
 export default class AVSService {
-  async getAll() {
-    const response = await apiGet('/avs');
+  async getAll(pageNumber) {
```
rubo, line 4: I'd expect this function to take the page size as well. So we can easily change it when needed.

vincenthongzy: Maybe when we have tables that allows them to choose row size? Currently the design is just fixed 10 rows per page

src/avs/AVSList.jsx
```diff
@@ -1,104 +1,144 @@
...
+    if (currentPage - 1 >= 1) {
+      setSearchParams({ page: (currentPage - 1).toString() });
+      fetchAVS(currentPage - 1);
+    }
+  };
+
+  const handlePageClick = page => {
+    setSearchParams({ page: page.toString() });
+    fetchAVS(page);
+  };
```
vincenthongzy, line 82: `nit`: `fetchAVS` and these should be wrapped in useCallback

vincenthongzy (approved):
are you going to implement the backend fix?

kushagrasarathe:
> are you going to implement the backend fix?

@vincenthongzy  not right now, I'm prioritising all the frontend changes currently, so that we can have it properly working for the MVP.

src/@services/AVSService.js
kushagrasarathe: @vincenthongzy  @rubo  so should I add this `page-size` param now or we do it later along with tables?

src/avs/AVSList.jsx
kushagrasarathe: fixed this 🫡

PR #104, opened 2024-07-04
Fix broken capitalization for LST icons

vincenthongzy:
MacOS problems, they treat files as case-insensitive. `git mv` to update the files properly


vincenthongzy:
image

nice

Zishan-7 (approved):

PR #112, opened 2024-07-08
AVS pages improvements

Zishan-7:


src/avs/AVSTotalValueOvertime.jsx
```diff
@@ -42,8 +44,15 @@ const AVSTotalValueOvertime = ({ avsAddress }) => {
       try {
         const avsTotalValueOvertimeData =
           await avsService.getAVSTotalValue(avsAddress);
+
+        const growthPercentage = getGrowthPercentage(
+          avsTotalValueOvertimeData[avsTotalValueOvertimeData.length - 2].tvl,
+          avsTotalValueOvertimeData[avsTotalValueOvertimeData.length - 1].tvl
+        );
```
vincenthongzy, line 51: this does not look like the correct parameters for growth percentage, we need to get the current slice's first and last values

vincenthongzy: e.g 
- for the 7 day slice, get first and last day
- for the 30 day slice ,get first and last day

vincenthongzy: image

vincenthongzy (changes requested):

src/avs/AVSTotalValueOvertime.jsx
Zishan-7: ah, didn't know this

src/avs/OperatorsOvertime.jsx
```diff
@@ -29,7 +29,14 @@ const OperatorsOvertime = ({ avsAddress }) => {
...
     if (!state.operatorsOvertimeData) return null;
-    return getDataByRange();
+    const data = getDataByRange();
+    dispatch({
+      growth: getGrowthPercentage(
+        data[0].operators,
+        data[data.length - 1].operators
+      )
+    });
+    return data;
```
vincenthongzy, line 39: I don't think this is intended, this should be in a useEffect that changes when filteredData change right?

vincenthongzy (changes requested):

src/avs/OperatorsOvertime.jsx
vincenthongzy: currently if u change the timeline tabs, the percentage doesn't change

Zishan-7: Its working fine at my end 🤔

PR #116, opened 2024-07-10
Add "Not found" page

rubo:
- Implemented "no match" route handler to display a dedicated page for not found URLs
- Revised layout to expand the content to screen size

### DIFF
--- a/src/shared/NotFound.jsx
+++ b/src/shared/NotFound.jsx
@@ -1,0 +1,14 @@
+export default function NotFound() {
+  return (
+    <div className="flex flex-col h-full">
+      <div className="font-display font-medium mb-4 text-foreground-1 text-3xl">
+        Not found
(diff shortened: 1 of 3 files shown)


src/shared/NotFound.jsx
vincenthongzy, line 1: image

on mobile view, does not happen with other pages

rubo: Fixed

vincenthongzy: the oversized twitter and discord logo will be fixed later yeah?

vincenthongzy (approved):

src/shared/NotFound.jsx
rubo: That's a good observation. Missed it somehow. Yes, next PR.

PR #117, opened 2024-07-11
AVS list page improvements

Zishan-7:
- Updated the List to use NextUI table
- Added search Filter
- Add sorting logic


src/avs/AVSList.jsx
```diff
@@ -107,62 +139,75 @@ export default function AVSList() {
...
+          loadingContent={<Spinner />}
+          emptyContent={
+            <span className="text-foreground-active">No AVS found.</span>
+          }
+        >
+          {state.avs?.map((avs, i) => (
+            <TableRow
+              onClick={() =>
+                navigate(`/avs/${avs.address}`, { state: { avs: avs } })
+              }
```
vincenthongzy, line 178: might be worthwhile to have a look? https://nextui.org/docs/components/table#why-not-array-map

vincenthongzy: nvm took a look at the interface, seems like its really annoying to adapt to it

src/avs/AVSList.jsx
```diff
@@ -191,76 +191,65 @@ export default function AVSList() {
...
+                >
+                  <TableCell className="p-4 flex gap-x-3">
+                    <span className="size-3">
+                      {(searchParams.get('page') - 1) * 10 + i + 1}
+                    </span>
+                    {avs.metadata?.logo ? (
+                      <img
+                        className="size-5 rounded-full"
+                        src={avs.metadata?.logo}
+                      />
```
vincenthongzy, line 231: image

let's add `bg-foreground-2` to make some icons like Witness Chain and Brevis more visible

vincenthongzy (changes requested):

src/shared/Layout.jsx
```diff
@@ -15,7 +15,7 @@ export default function Layout() {
             <Sidebar />
           </div>
         )}
-        <div className="flex flex-col flex-1">
```
vincenthongzy, line 18: why was this necessary? does changing this break any other page's layout?

src/avs/AVSList.jsx
```diff
@@ -1,52 +1,70 @@
...
 import { useTailwindBreakpoint } from '../shared/useTailwindBreakpoint';
-import { formatNumber } from '../utils';
+import useDebounce from '../shared/hooks/useDebounce';
+import { formatNumber } from '../shared/formatters';
+
+const columns = [
+  {
+    key: 'AVS',
+    label: 'AVS',
+    className: 'w-64 md:w-2/5'
```
vincenthongzy, line 25: OK i see that without this, on mobiles we get the table will all columns, then it expands into overflow and might not be a good UX. Will be good if you can also screenshot and comment so we can get better context on why these changes are necessary

vincenthongzy (changes requested):
text in table rows are not aligned , noticed that AVS name is not center aligned with the columns to its right

image

src/avs/AVSList.jsx
```diff
@@ -61,134 +79,176 @@ export default function AVSList() {
...
-        </div>
-        {state.isFetchingAvsData ? (
-          <AVSListSkeleton />
-        ) : (
-          <div>
-            {state.avs?.map(
-              (avs, i) =>
-                avs.metadata && (
-                  <div
+        <Table
```
vincenthongzy, line 160: please fix the warnings, missing `aria-label`

image

src/avs/AVSList.jsx
```diff
@@ -61,134 +79,176 @@ export default function AVSList() {
...
+                className={`bg-transparent py-4 text-foreground-active text-sm font-normal leading-5 ${column.className}`}
+                key={column.key}
+              >
+                {column.label}
+              </TableColumn>
+            )}
+          </TableHeader>
+          <TableBody
+            emptyContent={
+              <span className="text-foreground-active">No AVS found.</span>
```
vincenthongzy, line 180: should fix the pagination and disable everything

image

src/avs/AVSList.jsx
```diff
@@ -61,134 +79,176 @@ export default function AVSList() {
...
+                      <span className="size-3">
+                        {(searchParams.get('page') - 1) * 10 + i + 1}
+                      </span>
+                      {avs.metadata?.logo ? (
+                        <img
+                          className="size-5 rounded-full bg-foreground-2"
+                          src={avs.metadata?.logo}
+                        />
+                      ) : (
+                        <span class="material-symbols-outlined h-5 rounded-full text-lg text-yellow-300 min-w-5 flex justify-center items-center">
```
vincenthongzy, line 218:
```suggestion
                        <span className="material-symbols-outlined h-5 rounded-full text-lg text-yellow-300 min-w-5 flex justify-center items-center">
```

vincenthongzy (changes requested):
the search logic in the useEffect is a little buggy. Sometimes I search for non-existing address/names, it still shows me first page permanently or flash the first page then back to not found

https://github.com/user-attachments/assets/892bc647-d53b-4333-9178-4edf9a39bab2

src/shared/Layout.jsx
Zishan-7: This prevents the content overflowing the screen on mobile devices. This change does not break other page layout but fixes it by preventing unnecessary horizontal overflow

with `flex-1`

Screenshot 2024-07-16 at 1 18 10 PM

with `w-full`

Screenshot 2024-07-16 at 1 19 07 PM

src/avs/AVSList.jsx
Zishan-7: I have hidden the pagination component if table is empty

src/avs/AVSList.jsx
Zishan-7: We are using `layout= fixed` for the table. So to make the table scrollable in mobile devices we need to specify a fixed width instead of percentage so that the content overflows the table width and the scroll is enabled.
If we give the width in percentage the table will never scroll as it will try to fit into the available width

src/@services/AVSService.js
```diff
@@ -1,15 +1,29 @@
...
+
+  async getAll(pageNumber, search) {
     const pageIndex = Math.max(0, pageNumber - 1);
+
+    // Only abort if the request is pending and new request is fired
+    if (this.abortController) {
+      this.abortController.abort();
+    }
+    this.abortController = new AbortController();
+
```
Zishan-7, line 16: > the search logic in the useEffect is a little buggy. Sometimes I search for non-existing address/names, it still shows me first page permanently or flash the first page then back to not found. Please observe the behaviour when I search for `faj` multiple times
> 
>  Screen.Recording.2024-07-16.at.3.48.33.PM.mov

Added abort controller to counter the above mentioned bug. So if new search query is passed before the previous one was completed, it cancels the previous request and the results are only shown for the new one.

Zishan-7:
> text in table rows are not aligned , noticed that AVS name is not center aligned with the columns to its right
> 
> image

Fixed

src/@services/AVSService.js
vincenthongzy: the abort controller instantiation should be in the components useEffect

Zishan-7: Got it. Fixing this

rubo (changes requested):
Haven't yet reviewed the code closely, but there are a few things to address regarding the UI:

- The search input doesn't match the design. The border color.
- Pagination items are clickable `span` elements for some reason. Must be `Button` to show some interactivity.

src/shared/Layout.jsx
vincenthongzy: @rubo which pages are broken specifically, any screenshots? I've tested locally and it seems ok from my end. Most of the pages are currently broken on smaller screens, irregardless of this change

src/avs/AVSList.jsx
```diff
@@ -1,194 +1,273 @@
...
+    fetchAVS(params.page, params.search);
+    dispatch({ searchTriggered: false });
+  }, [searchParams, debouncedSearchTerm]);
+
+  useEffect(() => {
+    dispatch({ searchTriggered: true });
+  }, [debouncedSearchTerm]);
+
+  useEffect(() => {
+    if (state.sortDescriptor) console.log(state.sortDescriptor); // capture user's selection and recall the API with sorting params
```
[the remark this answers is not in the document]
Zishan-7: I have added it here as a dummy till we implement sorting.

src/shared/Pagination.jsx
```diff
@@ -17,12 +15,13 @@ const Pagination = ({
 
   return (
     <div className="flex border-t border-outline gap-x-2 justify-between items-center p-4 text-foreground-1 mt-austo">
-      <span
-        className={`cursor-pointer material-symbols-outlined ${activePage === 1 && 'text-disabled'}`}
+      <button
```
rubo, line 18: I mean the Button component of NextUI. BTW, please take a look at [the Pagination samples](https://nextui.org/docs/components/pagination). It's pretty good out-of-the-box. Why do we have it customized?

public/notFound.svg
rubo, line 1: We can drop this.

src/avs/AVSList.jsx
```diff
@@ -1,194 +1,298 @@
...
         <div className="font-display font-medium text-foreground-1 text-3xl">
           AVS
         </div>
-        <div className="font-display font-medium text-foreground-1 text-base">
-          Actively Validated Services
+        <div className="flex flex-col lg:flex-row gap-4 w-full justify-between lg:items-center items-start mb-6">
+          <div className="font-display font-medium text-foreground-1 text-base">
+            Actively Validated Services
+          </div>
+          <Input
```
rubo, line 167: The search box styles need to be updated as per the design.

rubo (changes requested):
Also, I'd like to apply the following changes to our search:

- Start searching only when the input has 3 or more characters. Typing just "a" doesn't make sense.
- If it starts with "0x", search when it has 5 (maybe 4) or more characters as "0xa" doesn't make sense.

The above helps to reduce the number of requests to the server.
Also, I'd like to change the address search pattern to match from the beginning. It's very uncommon to search for an address by its end or by something in the middle.
CC @vincenthongzy

vincenthongzy:
> Also, I'd like to apply the following changes to our search:
> 
> * Start searching only when the input has 3 or more characters. Typing just "a" doesn't make sense.
> * If it starts with "0x", search when it has 5 (maybe 4) or more characters as "0xa" doesn't make sense.
> 
> The above helps to reduce the number of requests to the server. Also, I'd like to change the address search pattern to match from the beginning. It's very uncommon to search for an address by its end or by something in the middle. CC @vincenthongzy

1. Agree. We need to update the search bar with a placeholder like "Start searching [AVS/Operator] address/name with 3 or more characters" or have an "error" text above the input field when we detect there's less than 3 characters and no more keystroke is registered. 

2. Depends on how we want the experience to be like though, if I don't prefix `0x` does it mean I'm never searching for addresses?  Otherwise we need to match anywhere for addresses. However most likely people will just copy and paste an address instead of manually typing it in, so we might be fine with a more rigid logic and just treat query without `0x` as names only?

Zishan-7:
> > Also, I'd like to apply the following changes to our search:
> > 
> > * Start searching only when the input has 3 or more characters. Typing just "a" doesn't make sense.
> > * If it starts with "0x", search when it has 5 (maybe 4) or more characters as "0xa" doesn't make sense.
> > 
> > The above helps to reduce the number of requests to the server. Also, I'd like to change the address search pattern to match from the beginning. It's very uncommon to search for an address by its end or by something in the middle. CC @vincenthongzy
> 
> 1. Agree. We need to update the search bar with a placeholder like "Start searching [AVS/Operator] address/name with 3 or more characters" or have an "error" text above the input field when we detect there's less than 3 characters and no more keystroke is registered so users don't get confused
> 2. Depends on how we want the experience to be like though, if I don't prefix `0x` does it mean I'm never searching for addresses?  Otherwise we need to match anywhere for addresses. However most likely people will just copy and paste an address instead of manually typing it in, so we might be fine with a more rigid logic and just treat query without `0x` as names only?

Should we add a tooltip next to search bar and mention our search criteria there?

rubo:
> Should we add a tooltip next to search bar and mention our search criteria there?

@Zishan-7 Yes, but you may skip it now to save time. Also, for now, we treat something as an address if it starts with "0x".

rubo (commented):
The AVS list now looks good. No major issues. We just need to add sorting. There are a few things to think about:

- Given the NextUI's Pagination is pretty powerful and well thought, I think it doesn't make sense to have left/right arrows in the bottom corners. The component already provides those in different ways.
- The hover color for the table header is gray -- we need to change it. If it's not straightforward, we can leave it for later, as there are some hover issues in NextUI.

Zishan-7:
> The AVS list now looks good. No major issues. We just need to add sorting. There are a few things to think about:
> 
> * Given the NextUI's Pagination is pretty powerful and well thought, I think it doesn't make sense to have left/right arrows in the bottom corners. The component already provides those in different ways.
> * The hover color for the table header is gray -- we need to change it. If it's not straightforward, we can leave it for later, as there are some hover issues in NextUI.

- I have changed the header hover color
- I will create a separate PR for adding sorting.

src/shared/hooks/useDebounce.jsx
```diff
@@ -1,17 +1,24 @@
-import { useState, useEffect } from 'react';
+import { useEffect, useState } from 'react';
 
-const useDebounce = (value, delay) => {
+const useDebounce = (value, delay, isSearchTerm) => {
```
vincenthongzy, line 3: Hm I don't like how the `useDebounce` signature now includes a very specific parameter that's only applicable for search bars.

We should keep this hook simple, and handle the validation at the table's useEffect hook before calling the API. So something like?:

```
if(validateSearchTerm) {
    avsService.getAll(...)
}
```

vincenthongzy (changes requested):
@rubo fix the linters other errors in a separate PR? Some of them can be fixed by deleting some unused files entirely after #118 is merged

src/shared/hooks/useDebounce.jsx
vincenthongzy: with this logic, we will need to update backend behaviour? So backend doesn't do an `$or` match anymore, but check if the query starts with `0x` then we match with `address` field, otherwise match `metadata.name` only?

```
// fe
  const validateSearchTerm = query => {
    const minQueryLength = query.startsWith('0x') ? 5 : 3;
    return query.length === 0 || query.length > minQueryLength;
  };

// be
           // update this so no need for `or` anymore?
            {
              $or: [
                { 'metadata.name': { $regex: new RegExp(search, 'i') } },
                { address: { $regex: new RegExp(search, 'i') } }
              ]
            }
```

Zishan-7: > with this logic, we will need to update backend behaviour? So backend doesn't do an $or match anymore, but check if the query starts with 0x then we match with address field, otherwise match metadata.name only?
Thanks for pointing this I will update the logic.

Zishan-7: > Hm I don't like how the useDebounce signature now includes a very specific parameter that's only applicable for search bars.

I added this directly in `useDebounce` to avoid writing the same `validateSearchTerm` logic in the components where search is needed assuming the criteria for search will be same across all other components.

vincenthongzy: The `useDebounce` hook should still be generic, so I would prefer moving it up and live with minimal duplication, or rename it to a more specific `useDebouncedSearch`

tailwind.config.js
```diff
@@ -8,7 +8,11 @@ export default {
     './node_modules/@nextui-org/theme/dist/**/*.{js,ts,jsx,tsx}'
   ],
   theme: {
-    extend: {},
+    extend: {
+      height: {
+        'table-skeleton-item': '3.82rem'
+      }
+    },
```
vincenthongzy, line 15: this shouldn't be inside tailwind config, because the height is specific for some pages only.

vincenthongzy (changes requested):

src/shared/hooks/useDebounce.jsx
vincenthongzy: @Zishan-7 does it run on your end?
image

Zishan-7: Renamed the hook to `useDebouncedSearch`

tailwind.config.js
Zishan-7: Got it, fixed.

vincenthongzy (approved):
We don't support a combination of sorting for now yeah? I.e desc stakers + asc TVL

src/avs/AVSList.jsx
```diff
@@ -1,194 +1,311 @@
...
+            ? 'descending'
+            : 'ascending'
+        }
+      : null
   });
 
+  const debouncedSearchTerm = useDebouncedSearch(
+    state.searchTerm ?? '',
+    300,
+    false
```
vincenthongzy, line 68: don't this false is necessary
```suggestion
```

src/shared/hooks/useDebouncedSearch.jsx
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react';
+
+const useDebouncedSearch = (value, delay) => {
+  const [debouncedValue, setDebouncedValue] = useState(value);
+
+  const validateSearchTerm = query => {
+    const minQueryLength = query.startsWith('0x') ? 5 : 3;
+    return query.length === 0 || query.length > minQueryLength;
```
vincenthongzy, line 8:
```suggestion
    return query.length === 0 || query.length >= minQueryLength;
```

src/avs/AVSList.jsx
```diff
@@ -1,194 +1,311 @@
...
+            }}
+            value={state.searchTerm ?? ''}
+            onChange={handleSearch}
+            type="text"
+            placeholder="Search by AVS"
+            radius="sm"
+            className="lg:w-96"
+            variant="bordered"
+            endContent={
+              <span className="material-symbols-outlined text-subtitle">
```
vincenthongzy, line 191: deprecating usage of `-subtitle`
```suggestion
              <span className="material-symbols-outlined text-foreground-2">
```

Zishan-7:
> We don't support a combination of sorting for now yeah? I.e desc stakers + asc TVL

No its just singe sorting field right now

rubo (approved):

PR #120, opened 2024-07-11
Consolidate number formatters

rubo:
Introduced `shared/formatters.js` that contains common number formatters we need:

- USD amount formatted as a rounded integer like $123,456 or compact $1.2B
- ETH amount formatted as a rounded integer like ETH 123,456 or compact like ETH 1.2B
- Number formatted as a rounded integer 123,456 or compact like 1.2B

Updated LRT view to use those.

### DIFF
--- a/src/lrt/LRTTotalValue.jsx
+++ b/src/lrt/LRTTotalValue.jsx
@@ -1,4 +1,4 @@
-import { fullNumFormatter } from './helpers';
+import { formatIntETH, formatIntUSD } from '../shared/formatters';
 import { Skeleton } from '@nextui-org/react';
 
 export default function LRTTotalValue({
@@ -20,13 +20,10 @@ export default function LRTTotalValue({
           {!isLoadingDelegations && (
             <div className="text-center">
               <div className="text-foreground-1 text-sm">
-                $
-                {fullNumFormatter.format(
-                  delegations?.total * delegations?.rate
-                )}
+                {formatIntUSD(delegations?.total * delegations?.rate)}
               </div>
               <div className="text-foreground-2 text-xs">
-                ETH {fullNumFormatter.format(delegations?.total)}
+                {formatIntETH(delegations?.total)}
               </div>
             </div>
           )}
--- a/src/lrt/LRTTotalValue.jsx
+++ b/src/lrt/LRTTotalValue.jsx
@@ -1,4 +1,4 @@
-import { formatIntETH, formatIntUSD } from '../shared/formatters';
+import { formatETH, formatUSD } from '../shared/formatters';
 import { Skeleton } from '@nextui-org/react';
 
 export default function LRTTotalValue({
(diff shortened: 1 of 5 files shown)


src/shared/formatters.js
```diff
@@ -0,0 +1,58 @@
...
+  return `${value / 1e3}k`;
+};
+
+// eslint-disable-next-line no-undef
+const ethIntFormatter = new Intl.NumberFormat('en-us', {
+  currency: 'ETH',
+  currencyDisplay: 'code',
+  maximumFractionDigits: 0,
+  style: 'currency'
+});
```
vincenthongzy, line 45: don't we also need compact formatters for the currency values?
```
const compactUSDFormatter = new Intl.NumberFormat('en-us', {
  currency: 'USD',
  currencyDisplay: 'symbol',
  maximumFractionDigits: 1,
  style: 'currency',
  notation: 'compact'
});
```

vincenthongzy: image

vincenthongzy (changes requested):

src/shared/formatters.js
rubo: This is a very good suggestion. Somehow missed that we can do compact formatting with the native formatter.

src/shared/formatters.js
```diff
@@ -1,40 +1,39 @@
 /**
- * Formats the given value as a rounded integer.
+ * Formats the given value as a rounded ETH amount.
  * @param {number | bigint} value The value to format.
+ * @param {boolean} compact Whether to use compact notation.
  * @returns {string} The formatted value.
  */
-export const formatInt = value => intFormatter.format(value);
+export const formatETH = (value, compact) =>
+  (compact ? ethCompactFormatter : ethIntFormatter).format(value);
```
vincenthongzy, line 8: This is a pretty neat trick, don't have to write out `format` twice. I didn't know we can do that with js

src/lrt/LRTTotalValue.jsx
```diff
@@ -20,10 +20,10 @@ export default function LRTTotalValue({
           {!isLoadingDelegations && (
             <div className="text-center">
               <div className="text-foreground-1 text-sm">
-                {formatIntUSD(delegations?.total * delegations?.rate)}
+                {formatUSD(delegations?.total * delegations?.rate)}
               </div>
               <div className="text-foreground-2 text-xs">
-                {formatIntETH(delegations?.total)}
+                {formatETH(delegations?.total)}
```
vincenthongzy, line 26: will you be passing in compact here for LRT components?

Previously it was done via `compact = !useTailwindBreakpoint('sm')`

vincenthongzy (approved):

src/lrt/LRTTotalValue.jsx
rubo: Well, we need compact formatting only when it's essential, such as in grids or chart axes. Otherwise, we display the full number when it fits. In this particular case, it fits the screen.

vincenthongzy: Clem didn't really like seeing the large numbers, they were of the opinion it's difficult to parse

PR #133, opened 2024-07-23
Implement AVS list error handling

Zishan-7:
- Added default sorting indication
- Added error handling
- Modified table empty state

This is how the errors will look:

Screenshot 2024-07-23 at 8 26 33 PM

When there is no AVS found for a search query the table headers will not be shown

Screenshot 2024-07-23 at 8 28 42 PM

### DIFF
--- a/src/avs/AVSList.jsx
+++ b/src/avs/AVSList.jsx
@@ -49,7 +49,7 @@ export default function AVSList() {
   const abortController = useRef(null);
   const [state, dispatch] = useMutativeReducer(reduceState, {
     avs: [],
-    isFetchingAvsData: false,
+    isFetchingData: false,
     searchTerm: searchParams.get('search'),
     error: null,
     rate: 1,
@@ -69,7 +69,7 @@ export default function AVSList() {
   const fetchAVS = useCallback(
     async (pageNumber, search, sort) => {
       try {
-        dispatch({ isFetchingAvsData: true, error: null });
+        dispatch({ isFetchingData: true, error: null });
 
         if (abortController.current) {
           abortController.current.abort();
@@ -85,7 +85,7 @@ export default function AVSList() {
 
         dispatch({
           avs: response.results,
-          isFetchingAvsData: false,
+          isFetchingData: false,
           rate: response.rate,
           totalPages: Math.ceil(response.totalCount / 10)
         });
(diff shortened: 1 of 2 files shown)


src/avs/AVSList.jsx
```diff
@@ -171,122 +172,131 @@ export default function AVSList() {
...
+
+      {!state.error && (
+        <div className="flex flex-1 flex-col rounded-lg border border-outline bg-content1 text-sm">
+          <Table
+            aria-label="AVS list"
+            layout="fixed"
+            removeWrapper
+            classNames={{
+              base: 'overflow-x-auto h-full',
+              table: 'h-full'
```
rubo, line 202: Not related to this PR, but I noticed a weird thing and couldn't find an "official" way to get rid of it. Changed the colors, so you can spot it:

image

This empty space is on each table. @vincenthongzy, you need to apply this to your tables as well.

```suggestion
              table: 'h-full',
              thead: '[&>tr:last-child]:hidden'
```

rubo (changes requested):
It shows the second screen with "No result found..." instead of an error. We need an error message regardless of whether it's a search or not.

The default sorting shows up fine.

src/avs/AVSList.jsx
Zishan-7: Added suggested change to the table component.

src/avs/AVSList.jsx
```diff
@@ -90,10 +91,10 @@ export default function AVSList() {
...
 
         abortController.current = null;
-      } catch (error) {
-        if (error.name !== 'AbortError') {
+      } catch (e) {
+        if (e.name !== 'AbortError') {
           dispatch({
-            error: 'Failed to fetch AVS data',
+            error: handleServiceError(e),
             isFetchingAvsData: false
```
rubo, line 98: Minor. Please follow the naming convention -- `isFetchingAVSData.` As this view loads only AVS data, we can keep it short `isFetching` or `isFetchingData`...

rubo (changes requested):
We made a mistake here. Please see https://github.com/NethermindEth/restaking-dashboard-backend/pull/95.

So, if there's an error, we display an error message using the `ErrorMessage` component as you have on the first screen. But, if AVS is not found, it returns an empty collection, and you show the second screen.

rubo (approved):

PR #132, opened 2024-07-22
Fix AVS list minor design issues

rubo:
Fixed some design issues on the AVS list view:

- Added border to AVS logos so they stand out more, and resized them according to the design
- Fixed some alignment issues
- For AVS with no metadata, display the address instead of the name and a question mark instead of a logo
- Revised and optimized page layout
- Removed hardcoded heights

Before
Screenshot 2024-07-22 191740
Screenshot 2024-07-22 192453

After
Screenshot 2024-07-22 191721
Screenshot 2024-07-22 192352

PR #136, opened 2024-07-24
Refactor and revise AVS views

rubo:
- Extracted AVS/operator logo to a separate component
- Revised AVS X link

### DIFF
--- a/src/shared/ThirdPartyLogo.jsx
+++ b/src/shared/ThirdPartyLogo.jsx
@@ -1,0 +1,23 @@
+import { Image } from '@nextui-org/react';
+
+/**
+ * @param {{ className: string, url: string }} props
+ */
+export default function ThirdPartyLogo({ className, url }) {
+  if (!url) {
+    return (
+      <span
+        className={`material-symbols-outlined flex items-center justify-center rounded-full bg-foreground-2 ${className}`}
+      >
+        question_mark
+      </span>
+    );
+  }
+
+  return (
+    <Image
+      className={`rounded-full border-2 border-foreground-2 bg-foreground-2 ${className}`}
+      src={url}
+    />
+  );
+}
--- a/src/avs/AVSDetailsHeader.jsx
+++ b/src/avs/AVSDetailsHeader.jsx
@@ -1,5 +1,6 @@
-import { Button, Image, Link, Skeleton } from '@nextui-org/react';
+import { Button, Link } from '@nextui-org/react';
 import { reduceState } from '../shared/helpers';
+import ThirdPartyLogo from '../shared/ThirdPartyLogo';
 import { useCallback } from 'react';
 import { useMutativeReducer } from 'use-mutative';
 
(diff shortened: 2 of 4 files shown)


Zishan-7 (approved):

src/avs/charts/TVLTabLineChart.jsx
```diff
@@ -149,7 +149,7 @@ export default function TVLTabLineChart({ points, height, width }) {
           <div className="font-display text-xl text-foreground-1">
             TVL over time
           </div>
-          <div className="flex gap-x-2 text-xs text-foreground-2">
+          <div className="flex gap-x-2 text-sm text-foreground-2">
```
vincenthongzy, line 152: I kept it `text-xs` because the subtitle in LRT chart is also `text-xs`. This is very confusing for me and giving extra work to reviewers to keep revising the styling, but I did so intentionally.

Moving forward, do I follow the design 100% or should I continue to adapt the design to what's already considered done (LRT page) so that everything is consistent?

src/shared/ThirdPartyLogo.jsx
vincenthongzy, line 1: image

if there is an url, but we fail to fetch the image its currently an empty space there. We should show a placeholder imo. Two solutions:

1. Keep the original solution of manually checking for errors and rendering a skeleton
2. Find a fallback image and use `fallbackSrc` prop of the `Image` component.

src/shared/helpers.js
```diff
@@ -3,3 +3,6 @@ export const reduceState = (draft, action) => void Object.assign(draft, action);
 export const handleServiceError = e => ({
   message: e?.message || 'An unknown error occurred'
 });
+
+export const truncateAddress = address =>
+  `${address.slice(0, 6)}...${address.slice(-4)}`;
```
vincenthongzy, line 8: what is the criteria of creating this `truncateAddress` function and moving it to `helpers.js` when sometimes in the reviews some helper functions inside components are being deleted to favour for inlining?

Example in #132 the truncate function was deleted and moved inline. Is this better housed in the helpers file too?
image

src/avs/charts/TVLTabLineChart.jsx
rubo: I also noticed that. But here the title is another font, it's bigger, so the subtitle should also be bigger. Anyway, we have to ask Carlos to unify the design as it's confusing as you said.

src/shared/helpers.js
rubo: My assumption is that address is a common thing and may be used somewhere else as well, like in operators. So I made it shared. The truncation function above may also be made shared, when you start with operators. Expect such back and forth until we finish the refactoring. So when when we see we have room for improvements or need revision, we do it.

src/shared/ThirdPartyLogo.jsx
rubo: Skeleton is an indication that something is loading. We need a fallback image or a so solid color. As it's a separate component now, we can address that later for everything.

vincenthongzy (approved):

PR #138, opened 2024-07-24
Move colors to config file

rubo:
Moved some hardcoded colors to `tailwind.config.js`.


tailwind.config.js
```diff
@@ -20,45 +20,49 @@ export default {
...
-            focus: '#34d399',
-            foreground: '#000000',
-            'dark-blue': '#009CDD'
-          }
+          colors: {}
         },
         dark: {
           colors: {
             background: '#080a0f',
+            chart: {
```
vincenthongzy, line 28: we need 2 more colors for LST chart

vincenthongzy (changes requested):

tailwind.config.js
rubo: I don't have them now. Will add later.

vincenthongzy (approved):

PR #143, opened 2024-07-25
Revise Home view

rubo:
- [x] TVL box
- [x] Top AVS
- [x] Top operators
- [x] EigenLayer TVL chart
- [x] LRT pie chart


src/index.css
```diff
@@ -16,7 +16,7 @@ body {
 }
 
 .rd-box {
-  @apply rounded-lg border border-outline bg-content1;
+  @apply overflow-hidden rounded-lg border border-outline bg-content1;
```
vincenthongzy, line 19: Why do we need overflow-hidden here? Currently our `Table`s are wrapped with `<div className='rounded-lg border border-outline bg-content1">`, and if we migrate to `.rd-box` , I'm not sure if it will cause our horizontally scrollable table to be unscrollable

rubo: We need it to clip container's content to its borders as borders are rounded. Table's scroll is its own scroll, so it works and can be easily checked.

Zishan-7 (approved):

src/index.css
rubo: @vincenthongzy This is for you to get a better understanding of why's the change:

image

PR #144, opened 2024-07-25
Revamp "Not Found" view

rubo:
Revamped and improved the "Not Found" view

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -38,6 +38,7 @@ export default [
         }
       ],
       'react/jsx-no-target-blank': 'off',
+      'react/no-unescaped-entities': 'off',
       'react/prop-types': 'off',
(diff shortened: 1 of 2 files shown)


Zishan-7 (approved):

eslint.config.js
```diff
@@ -38,6 +38,7 @@ export default [
         }
       ],
       'react/jsx-no-target-blank': 'off',
+      'react/no-unescaped-entities': 'off',
```
vincenthongzy, line 41: think its a good opportunity to also add this in so we can automatically sort props alphabetically?

'react/jsx-sort-props': 'warn',

vincenthongzy (approved):

eslint.config.js
rubo: Applied

PR #148, opened 2024-07-26
Fix AVS Details tab navigation

### LINKED ISSUE #137: Navigation misbehavior on AVS details view
When you go to AVS details operator tab and then try to go back to the previous page, it does not work. it just gets stuck there.
 
 

https://github.com/user-attachments/assets/9ce0ebcd-98c9-41db-9c19-6d308a916c52

vincenthongzy:
Fixes #137 

1. Now changing tabs will update the url path with an extra path parameter. This parameter is optional.
- old : /avs/:address
- new: /avs/:address/tvl & /avs/:address/operators

2. To keepthings sane, we keep two entries on the history stack:

- /tvl tab
- /operators tab <--- this gets replaced when search params change instead of pushing new entries

3. Now users will just paste this link to directly access the table
http://localhost:5173/avs/0x870679e138bcdf293b7ff14dd44b70fc97e12fc0/operators?page=1&sort=-tvl


src/App.jsx
```diff
@@ -18,33 +18,29 @@ import OperatorDetails from './operators/OperatorDetails';
...
+      <Route path="/avs" element={<AVSList />} />
+      <Route path="/avs/:address/:tab?" element={<AVSDetails />} />
+      <Route path="/lrt" element={<LRT />} />
+      <Route path="/operators" element={<OperatorsList />} />
+      <Route path="/operators/:address" element={<OperatorDetails />} />
+      <Route path="/lst" element={<LST />} />
+      <Route path="*" element={<NotFound />} />
+    </Route>
+  )
+);
```
vincenthongzy, line 34: moved this up as per https://reactrouter.com/en/main/routers/router-provider. I need to access the router instance

src/App.jsx
```diff
@@ -18,33 +18,29 @@ import OperatorDetails from './operators/OperatorDetails';
...
 
 export default function App() {
   log.debug('Starting up');
 
   return (
-    <NextUIProvider>
+    <NextUIProvider navigate={router.navigate}>
```
vincenthongzy, line 40: So that the `href` in Tab can work properly.  The docs say this is "PRIVATE - DO NOT USE" , but its perfectly fine to use as per the [maintainer](https://github.com/remix-run/react-router/issues/9422#issuecomment-1301182219). 

If we are not feeling comfortable with using this "unstable" API, we would have to remove the usage of `createBrowserRouter` and `RouterProvider` and use <BrowserRouter/> instead. From what I can tell, we don't use any of the fancy data API features provided by `RouterProvider`. More info [here](https://nextui.org/docs/guide/routing#react-router)

rubo (approved):
Good job!

PR #152, opened 2024-07-28
feat: graph loading skeleton 

kushagrasarathe:
- added a graph skeleton component to show while data is being fetched

> Note: @rubo the component is only added in the LST page currently to demo it, will update it on rest of the places once you approve it 

https://github.com/user-attachments/assets/7df7ea75-f1fe-4f52-b4f2-ddbd6e182437


rubo:
Will revisit this in the future.

PR #96, opened 2024-06-20
Feature/add prod env

jianmtech:

PR #159, opened 2024-07-31
Refactor overview LRT distribution

Zishan-7:
- Remove scroll from Top AVS and top operators table
- Refactor LRT distribution component

Need to merge [#96](https://github.com/NethermindEth/restaking-dashboard-backend/pull/96) before merging this one
Also, please checkout to `fix/usd-rate-latest-lrt-api` branch on the backend while checking this locally


src/home/LRTDistribution.jsx
```diff
@@ -0,0 +1,183 @@
...
+          lrt.tvl = results.protocols[p];
+          lrtTVL += results.protocols[p];
+          lrtDistribution.push(lrt);
+        });
+
+        dispatch({
+          lrtDistribution: lrtDistribution
+            .sort((a, b) => b.tvl - a.tvl)
+            .filter(lrt => (lrt.tvl / lrtTVL) * 100 > 1), // Filter and display only LRTs that have more than 1% share
+          lrtTVL,
```
Zishan-7, line 44: Showing only the LRTs which have more than 1% of share because the values less than 1% don't display correctly on the pie chart.
Screenshot 2024-07-31 at 6 53 15 PM

rubo (changes requested):

src/home/LRTDistribution.jsx
rubo: I think it's ok. We can't do much here.

src/lrt/helpers.js
```diff
@@ -10,3 +10,46 @@ export const protocols = {
   renzo: { index: 2, name: 'Renzo' },
   swell: { index: 5, name: 'Swell' }
 };
+
+export const lrtProtocolsMapping = {
```
rubo, line 14: I meant putting these values in the mapping above. Actually, just `logo`. Otherwise, we duplicate the data.

src/home/charts/LRTDistributionPieChart.jsx
```diff
@@ -0,0 +1,132 @@
...
+            <Text dy={0} fill="hsl(var(--app-foreground))" textAnchor="middle">
+              {formatUSD(lrtTVL * ethRate, true)}
+            </Text>
+            <Text
+              className="text-sm"
+              dy={20}
+              fill="hsl(var(--app-foreground-1))"
+              textAnchor="middle"
+            >
+              {formatETH(lrtTVL, true)}
```
rubo, line 117: These are still compact. We have room for them.

rubo (changes requested):

src/home/LRTDistribution.jsx
```diff
@@ -0,0 +1,183 @@
...
+            <ThirdPartyLogo className="size-5 min-w-5" url={logo} />
+            {label}
+          </div>
+        }
+        radius="sm"
+        showValueLabel={true}
+        value={value}
+      />
+      <div className="ml-2 min-w-fit text-sm text-foreground-2">
+        {formatNumber(tvl, true)} ETH
```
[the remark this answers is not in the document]
rubo: Perhaps you forgot.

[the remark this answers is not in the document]
Zishan-7: Oh, I added compact numbers because in the current implementation we had compact numbers here. I have updated it in the new implementation.

src/lrt/helpers.js
Zishan-7: Got it. I have updated it

rubo (approved):

PR #161, opened 2024-08-01
Refactor EigenLayer TVL overtime chart

Zishan-7:


src/home/charts/EigenLayerTVLOvertimeChart.jsx
```diff
@@ -0,0 +1,482 @@
...
+    <div className="rd-box min-h-44 basis-full p-4" ref={rootRef}>
+      <div className="mb-6 flex flex-wrap items-start justify-end gap-2 md:items-center">
+        <div className="flex-1">
+          <div className="text-base text-foreground-1">TVL over time</div>
+          <div className="text-sm text-foreground-2">{getLatestTotal}</div>
+        </div>
+        <Tabs
+          classNames={tabs}
+          defaultSelectedKey="eth"
+          disabledKeys={['usd']}
```
Zishan-7, line 244: We don't have USD rates in eigenlayer tvl API rn, so I have disabled the USD tab for now

rubo (changes requested):

src/home/charts/EigenLayerTVLOvertimeChart.jsx
```diff
@@ -0,0 +1,482 @@
...
+        value += parseFloat(BigInt(d.ethTVL) / BigInt(1e18));
+        value += parseFloat(BigInt(d.lstTVL) / BigInt(1e18));
+
+        return { timestamp: new Date(d.timestamp).getTime(), value };
+      }),
+      brushPosition: {
+        start: scaleBrushDate(
+          new Date(
+            eigenLayerTVL[eigenLayerTVL.length - 1 - 90].timestamp
+          ).getTime()
```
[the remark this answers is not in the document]
Zishan-7: In the other stacked area chart APIs we get unix timestamp but in the eigenlayer chart api we get utc string. So, I converted the UTC string to unix timestamp wherever we are dealing with timestamps. I tried plotting the graph with UTC string timestamps but it wasn't working

Screenshot 2024-08-01 at 11 44 52 PM
Screenshot 2024-08-01 at 11 45 17 PM

src/home/charts/EigenLayerTVLOvertimeChart.jsx
Zishan-7: We can add it. Until we do I have disabled the tab so that it does not break the page

src/home/charts/EigenLayerTVLOvertimeChart.jsx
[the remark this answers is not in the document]
rubo: Alright, let's fix the backend API first, then merge this PR. We need USD rate and fix timestamps. Let me know whether you need a help with it.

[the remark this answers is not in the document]
vincenthongzy: @Zishan-7 if the backend gets fixed, I need to update the LST page as well, please do give me a ping once the change is made

src/home/Home.jsx
```diff
@@ -41,7 +42,11 @@ export default function Home() {
         eigenLayerTVLError={state.error}
         isFetchingEigenlayerTVL={state.isFetchingEigenLayerTVL}
       />
-      <div className="rd-box min-h-44 basis-full"></div>
+      <EigenLayerTVL
+        error={state.error}
+        isFetching={state.isFetchingEigenLayerTVL}
+        tvl={state.eigenLayerTVL}
+      />
```
vincenthongzy, line 49: might be better not passing in `error` and `isFetching` and just handle these states at the `Home` level?

Zishan-7: imo its better to let the component handle its error and loading states. 🤔

vincenthongzy: The reasoning it because it's not EigenLayerTVL itself that makes the API call. Can we move the API call into the component instead?

Zishan-7: Adding the API call in the parent component saves us 1 API call because the data coming from `/eigenlayer/tvl` is required in two components `<EigenLayerTVL/>` and `<OverviewStats/>`

rubo (approved):

PR #162, opened 2024-08-02
Refactor LST page

vincenthongzy:


tailwind.config.js
```diff
@@ -32,10 +32,12 @@ export default {
...
-              7: '#ec4899',
-              8: '#f9a8d4',
-              9: '#0ea5e9',
-              10: '#7dd3fc'
+              7: '#ffcc80',
+              8: '#ec4899',
+              9: '#f9a8d4',
+              10: '#0ea5e9',
+              11: '#7dd3fc',
+              12: '#4fc3f7'
```
vincenthongzy, line 40: added 1 more shade of yellow `ffcc80` and 1 more shade of blue `4fc3f7`

src/lst/LSTList.jsx
```diff
@@ -1,82 +1,80 @@
...
-                    className="h-5 min-w-5 rounded-full bg-contain bg-center bg-no-repeat"
-                    style={{
-                      backgroundImage: `url('${STRATEGY_ASSET_MAPPING[address].logo}')`
-                    }}
-                  ></div>
-                  <div className="basis-full truncate">
-                    <span className="text-subtitle">
-                      {STRATEGY_ASSET_MAPPING[address].name}
+    <div className="rounded-lg border border-outline bg-content1 text-sm">
+      <Table
```
vincenthongzy, line 18: Different from LRT, opted to use Table here because easier to get the correct widths for the columns

src/lst/LSTDistributionGraph.jsx
```diff
@@ -411,28 +382,27 @@ export default function LSTDistributionGraph({
             {tooltipDateFormatter.format(new Date(tooltipData.x))}
           </div>
           <ul className="text-sm">
-            {state.keys?.map((key, i) => {
+            {state.keys.map((key, i) => {
```
vincenthongzy, line 385: @rubo I noticed that in LRT page you sort based on `tooltipData` but this seems to have slight issues with the color legends.

In the LRTList below , `Swell` could be yellow color. But in the tooltip depending on the data point, `Swell` could be any of the viable chart colors depending on the sorting order on that day

tailwind.config.js
vincenthongzy: oopd looks like I broke the color of the line charts

vincenthongzy: fixed

Zishan-7:
The graph behaves a bit weirdly for me: 

- On page load, graph shows all the data but the brush shows 3 months timeline

Screenshot 2024-08-02 at 5 00 00 PM

- If i move the brush to the left side it overflows and shows up over the actual graph
Screenshot 2024-08-02 at 5 00 35 PM

src/lst/LSTDistributionGraph.jsx
```diff
@@ -189,80 +174,75 @@ export default function LSTDistributionGraph({
         return { timestamp: d.timestamp, value };
       }),
       brushPosition: {
-        start: scaleBrushDate(data[0]?.timestamp),
-        end: scaleBrushDate(data[data.length - 1]?.timestamp)
+        start: scaleBrushDate(points[points.length - 1 - 90].timestamp),
+        end: scaleBrushDate(points[points.length - 1].timestamp)
       },
+      filterPoints: points.slice(-90),
```
Zishan-7, line 180:
```suggestion
      filteredPoints: points.slice(-90),
```
This will fix the brush and graph timeline mismatch on page load as well

Zishan-7 (changes requested):

src/lst/LST.jsx
```diff
@@ -1,128 +1,101 @@
...
 
     fetchLSTTotalValue();
-  }, []);
+  }, [dispatch, eigenlayerService]);
 
   return (
     <div>
-      <div className="my-4 font-display text-3xl font-medium text-foreground-1">
+      <div className="mb-4 font-display text-3xl font-medium text-foreground-1">
         Liquid Staking Tokens
```
rubo, line 63: Capitalization must be like everywhere else:

```suggestion
        Liquid staking tokens
```

src/lst/LSTList.jsx
```diff
@@ -1,82 +1,80 @@
...
+              <TableRow
+                className="border-t border-outline"
+                key={`lst-item-${i}`}
+              >
+                <TableCell>
+                  <div className="flex items-center gap-x-2">
+                    <span className="inline-block min-w-4 text-foreground-2">
+                      {i + 1}
+                    </span>
+                    <span
```
rubo, line 51: We need to show here the logos as it was previously and as we have it on the design. I'll commit the updated LST logos to this branch later.

rubo (changes requested):
- Tooltip title padding is missing
- The pointer indicator vertical line is misplaced
- The brush colors are wrong

image

This PR definitely isn't ready for a review. Making it draft. Please mark it as ready when the obvious issues are addressed.

src/lst/LSTDistributionGraph.jsx
vincenthongzy: I was looking at an old implement in LRTDistribution when I should have looked at LRTList instead, hence the confusion.

vincenthongzy: Anyways the problem is we shouldn't sort the day's data in the graph because it will cause the color legends to be incorrect. Notice Eigenpie and Kelp colors are swapped. This is because the whole graph is sorted by the latest day's data 

```
      keys: Object.entries(data?.[data.length - 1].protocols)
        .sort(sortProtocols)
        .map(([k]) => k)
```

But in LRT's tooltip we are sorting by each day's data again, causing the color legends to be incosistent with the overall graph.


image


image

Zishan-7 (approved):

tailwind.config.js
rubo: We just need 2 extra colors here. No need to modify already existing ones as it badly affects other parts of UI.

rubo (changes requested):

rubo (approved):
Colors revised.

src/lst/LSTDistributionGraph.jsx
rubo: This is a very good catch! Fixed in https://github.com/NethermindEth/restaking-dashboard/pull/163/commits/8eab7bbce1e868382af3345f58a888ff4515da30 so the tooltip is always sorted and doesn't display anything not on the graph for a given point unlike LST. Thus, LST chart needs to be changed that way. However, both approaches don't handle the case where protocol may be in the past but not in the present.

PR #177, opened 2024-08-14
Adjust all graphs to default to 'All' timeline

vincenthongzy:
https://www.notion.so/nethermind/Default-graph-time-range-to-All-for-more-visually-different-graph-8bf8940b6a7448efb6b14c4081d82977?pvs=4


Zishan-7 (approved):

rubo (changes requested):
- Refactored LRT/LST approach to have a single point of control. Using `ParentSize` is problematic as it fires lots of negative values.
- All other changes have no effect as setting the default selected key doesn't fire the handler. AVS TVL is a fixed sample suggestion.

rubo:
I'm not sure changing the default time range is a good idea at all. The more time passes, the less sense it makes.

src/avs/charts/OperatorsTabLineChart.jsx
```diff
@@ -264,7 +268,8 @@ export default function OperatorsTabLineChart({ points, height, width }) {
   );
 }
 
-const margin = { top: 20, right: 40, bottom: 24, left: 0 };
+const margin = { top: 20, right: 40, bottom: 30, left: 20 };
```
vincenthongzy, line 271: bumped the margins for line charts, otherwise the leftmost date label on 1w graphs gets gobbled up

vincenthongzy: image

Notice `Aug` also gets cut off

Zishan-7 (approved):

src/operators/charts/OperatorRestakersLineChart.jsx
```diff
@@ -342,7 +346,8 @@ function LineChart({ points, height, width }) {
   );
 }
 
-const margin = { top: 20, right: 40, bottom: 24, left: 0 };
+const margin = { top: 20, right: 40, bottom: 30, left: 20 };
```
rubo, line 349:
```suggestion
const margin = { top: 20, right: 40, bottom: 30, left: 0 };
```

src/operators/charts/OperatorTVLLineChart.jsx
```diff
@@ -373,7 +377,8 @@ function LineChart({ points, height, width }) {
   );
 }
 
-const margin = { top: 20, right: 40, bottom: 24, left: 0 };
+const margin = { top: 20, right: 40, bottom: 30, left: 20 };
```
rubo, line 380:
```suggestion
const margin = { top: 20, right: 40, bottom: 30, left: 0 };
```

PR #179, opened 2024-08-14
Fix AVS details minor bugs

vincenthongzy:
[RESTAKE-181](https://www.notion.so/nethermind/NaN-display-in-token-distribution-f0555c657c8b434396c1fa71f65bb146?pvs=4)

[RESTAKE-182](https://www.notion.so/nethermind/Empty-state-for-LST-Distribution-Heat-Map-677960d0aaae4946904e6e40c296eca2?pvs=4)

image


vincenthongzy:
There is another issue, https://www.notion.so/nethermind/Total-operators-count-doesn-t-match-list-numbering-287b83b103084af8b201fcc19a0fb460?pvs=4. I think this is not really fixable, it's probably due to the gap between avs indexer and operator indexer?

rubo (changes requested):
The "No distribution message" must not be an error message. It's not an error. Better to copy the "not found" message styling from AVS list.

Zishan-7 (approved):

src/avs/charts/TVLTabTreemap.jsx
```diff
@@ -57,6 +57,11 @@ export default function TVLTabTreemap({ width, height, ethRate, lst }) {
     [children]
   );
 
+  const isEmpty = useMemo(
+    () => root.children.reduce((acc, child) => acc + child.data.value, 0) === 0,
```
rubo, line 61: It throws here for AVS `0x18343aa10e3d2f3a861e5649627324aead987adf`:

> TypeError: Cannot read properties of undefined (reading 'reduce')

vincenthongzy: I am not able to reproduce locally when running on staging DB, but I've added in a guard to prevent this from happening

src/avs/AVSDetailsOperatorsTab.jsx
```diff
@@ -352,7 +352,11 @@ function AVSOperatorsList({ address, avsError, isAVSLoading, tvl }) {
                     </div>
                   </TableCell>
                   <TableCell className="pe-8 text-end">
-                    <div>{((op.strategiesTotal / tvl) * 100).toFixed(2)}%</div>
+                    <div>
+                      {tvl === 0
+                        ? 'N/A'
+                        : `${((op.strategiesTotal / tvl) * 100).toFixed(2)} %`}
```
rubo, line 358:
```suggestion
                        : `${((op.strategiesTotal / tvl) * 100).toFixed(2)}%`}
```

rubo (approved):

PR #180, opened 2024-08-14
Add LRT/LST tooltip total

rubo:
Implemented total calculation in LRT and LST tooltip.


src/lst/LSTDistribution.jsx
vincenthongzy, line 1: need to update the import path in `src/lst/LST.jsx`

vincenthongzy (changes requested):

src/lst/LSTDistribution.jsx
rubo: Whoops, Vite managed to handle that without error until I restarted it.

vincenthongzy (approved):

PR #182, opened 2024-08-14
Add search hint tooltip

rubo:
Added hint tooltip to search inputs.

### DIFF
--- a/src/shared/SearchTooltip.jsx
+++ b/src/shared/SearchTooltip.jsx
@@ -1,0 +1,36 @@
+import { Tooltip } from '@nextui-org/react';
+import { tooltip } from './slots';
+import { useState } from 'react';
+
+export default function SearchTooltip() {
+  const [isOpen, setOpen] = useState(false);
+
+  return (
+    <Tooltip
+      classNames={tooltip}
+      content={
+        <ul className="list-inside list-disc">
+          <li>Enter 3 or more more characters to start searching</li>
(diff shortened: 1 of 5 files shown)


src/shared/SearchTooltip.jsx
```diff
@@ -0,0 +1,36 @@
...
+  const [isOpen, setOpen] = useState(false);
+
+  return (
+    <Tooltip
+      classNames={tooltip}
+      content={
+        <ul className="list-inside list-disc">
+          <li>Enter 3 or more more characters to start searching</li>
+          <li>
+            Searches starting with "<code>0x</code>" will search for matching
```
vincenthongzy, line 15: do we need to specify that for addresses we only start searching when there are 5+ characters?

vincenthongzy (approved):

Zishan-7 (approved):

src/shared/SearchTooltip.jsx
rubo: Do we do that when 5+? If so, then yes

vincenthongzy: yeah frontend validation works that way in `useDebouncedSearch`

```
  const validateSearchTerm = query => {
    const minQueryLength = query.startsWith('0x') ? 5 : 3;
    return query.length === 0 || query.length >= minQueryLength;
  };
```

PR #187, opened 2024-08-15
Fix search debounce

rubo:
Fixed AVS search debounce not to make many requests in a row when search parameters changed. This has been caused by incorrect dependency configuration in `useEffect`. Therefore, `useEffect` is now split into 2 different ones with different set of dependencies.

The fix should be applied to other views as well.

### DIFF
--- a/src/avs/AVSList.jsx
+++ b/src/avs/AVSList.jsx
@@ -54,7 +54,6 @@ export default function AVSList() {
     searchTerm: searchParams.get('search'),
     error: null,
     rate: 1,
-    searchTriggered: false,
     sortDescriptor: searchParams.get('sort')
       ? {
           column: searchParams.get('sort').replace('-', ''),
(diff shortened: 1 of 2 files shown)


src/avs/AVSList.jsx
```diff
@@ -107,44 +107,49 @@ export default function AVSList() {
 
   const handlePageClick = useCallback(
     page => {
-      setSearchParams({ page: page.toString() });
+      searchParams.set('page', page.toString());
```
rubo, line 110: Do we need to copy it before mutating?

Zishan-7 (changes requested):
The multiple API requests bug is fixed but now if I change the page and try to go back using browser's back button, I have to click 2 times to return to the previous page.

https://github.com/user-attachments/assets/41036e79-aff8-4573-877e-e31471fb05d6

vincenthongzy:
@rubo does the same behaviour happen for AVS details operator table? I implemented the search functionality differently there

rubo:
@vincenthongzy I checked; it works fine. Maybe we can apply that approach here as well to avoid issues with navigation.

PR #188, opened 2024-08-16
Add AVS list to Operator details

vincenthongzy:
After discussing with Clem, simple list for now.

### DIFF
--- a/src/avs/AVSDetails.jsx
+++ b/src/avs/AVSDetails.jsx
@@ -105,7 +105,7 @@ export default function AVSDetails() {
           cursor: 'rounded border border-outline shadow-none',
           panel: 'p-0',
           tab: 'h-fit p-2',
-          tabList: 'rd-box w-full p-2'
+          tabList: 'rd-box w-full !overflow-x-scroll p-2'
         }}
         disabledKeys={['restakers']}
         radius="sm"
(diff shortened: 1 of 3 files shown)


src/avs/AVSDetails.jsx
```diff
@@ -105,7 +105,7 @@ export default function AVSDetails() {
           cursor: 'rounded border border-outline shadow-none',
           panel: 'p-0',
           tab: 'h-fit p-2',
-          tabList: 'rd-box w-full p-2'
+          tabList: 'rd-box w-full !overflow-x-scroll p-2'
```
vincenthongzy, line 108: @rubo the `rd-box`s overflow hidden caused the tab list horizontal scroll to stop working on smaller screens

rubo: It works now on a mobile. What's the problem?

vincenthongzy: It fails when there are errors. The error messages causes each tab to be longer than usual, causing tabs to be cut off.

rubo: Got it. I'll check but for this particular tab disabling it is ok.

Zishan-7 (approved):

src/avs/AVSDetails.jsx
```diff
@@ -122,7 +122,7 @@ export default function AVSDetails() {
               </div>
               <div className="flex w-full justify-center">
                 {state.isAVSLoading && (
-                  <Skeleton className="mt-2 h-4 w-full rounded-md" />
+                  <Skeleton className="mt-2 h-4 w-full min-w-16 rounded-md" />
```
rubo, line 125: For short texts like "TVL", skeleton expansion is too short. So, added a min width.

rubo (approved):

PR #197, opened 2024-08-24
change subdomain of stage env

PhilexWong:


.env.production
vincenthongzy, line 1: i don't think this is the right file, should create `.env.stage` with the same contents imo.

rubo:
There's already another PR fir this. Closing in favor of #196.

PR #200, opened 2024-08-27
add staging env config and amplify.yml

PhilexWong:
add amplify.yaml to support dynamic build with different backend endpoints for different environments.

 'if [[ "$AWS_BRANCH" == release* ]]; then npm run build:production; 
elif [[ "$AWS_BRANCH" == "main" ]]; then npm run build:staging; 
else npm run build:development; fi'

### DIFF
--- a/.env.staging
+++ b/.env.staging
@@ -1,2 +1,0 @@
-VITE_API_BASE_URL=https://api-stage.restaking.info
-VITE_LOG_LEVEL=info
--- a/.env.staging
+++ b/.env.staging
@@ -1,0 +1,2 @@
+VITE_API_BASE_URL=https://api-stage.restaking.info
+VITE_LOG_LEVEL=info
--- a/amplify.yml
+++ b/amplify.yml
@@ -6,7 +6,9 @@ frontend:
                 - 'npm ci --cache .npm --prefer-offline'
         build:
             commands:
-              - 'if [[ "$AWS_BRANCH" == release* ]]; then npm run build:production; elif [[ "$AWS_BRANCH" == "main" ]]; then npm run build:staging; else npm run build:development; fi'
+              - 'if [[ "$AWS_BRANCH" == "main" ]]; then npm run build:staging; fi'
+              - 'if [[ "$AWS_BRANCH" == release* ]]; then npm run build:production; fi'
+              - 'if [[ "$AWS_BRANCH" != "main" && "$AWS_BRANCH" != release* ]]; then npm run build; fi'
     artifacts:
         baseDirectory: dist
         files:
--- a/amplify.yml
+++ b/amplify.yml
@@ -6,9 +6,7 @@ frontend:
                 - 'npm ci --cache .npm --prefer-offline'
         build:
             commands:
-              - 'if [[ "$AWS_BRANCH" == "main" ]]; then npm run build:staging; fi'
-              - 'if [[ "$AWS_BRANCH" == release* ]]; then npm run build:production; fi'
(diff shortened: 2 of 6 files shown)


amplify.yml
```diff
@@ -0,0 +1,16 @@
+version: 1
+frontend:
+    phases:
+        preBuild:
+            commands:
+                - 'npm ci --cache .npm --prefer-offline'
+        build:
+            commands:
+              - 'if [[ "$AWS_BRANCH" == release* ]]; then npm run build:production; elif [[ "$AWS_BRANCH" == "main" ]]; then npm run build:staging; else npm run build:development; fi'
```
vincenthongzy, line 9: `nit`, not a blocker but personally i prefer the formatting in the [docs](https://docs.aws.amazon.com/amplify/latest/userguide/build-settings.html):

```
frontend:
  phases:
    build:
      commands:
        - if [ "${AWS_BRANCH}" = "main" ]; then echo "main branch"; fi
        - if [ "${AWS_BRANCH}" = "dev" ]; then echo "dev branch"; fi
```

amplify.yml
```diff
@@ -0,0 +1,16 @@
...
+    artifacts:
+        baseDirectory: dist
```
vincenthongzy, line 11: shouldnt this be `public` ?

amplify.yml
vincenthongzy, line 1: if we are using this workflow, i think we need to delete`deploy-stage.yml` ? otherwise seems like its a conflict

PhilexWong: There is no conflict. The git action in deploy-stage.yml is just a webhook to trigger to run amplify.yaml. 
amplify.yaml will be executed in amplify aws to deploy frontend.

vincenthongzy: Thanks, learned something new today!

amplify.yml
PhilexWong: This file, amplify.yml, is a public file, no confidential info. IMO, it can be moved to Amplify, but keeping it in the Git repository makes it much clearer and more straightforward. 

image

amplify.yml
PhilexWong: OK. I will change it as your suggestion.

amplify.yml
vincenthongzy: no i mean:

```suggestion
        baseDirectory: public
```

because without this our React frontend won't work properly since its missing our images directory

amplify.yml
```diff
@@ -0,0 +1,18 @@
...
+frontend:
+    phases:
+        preBuild:
+            commands:
+                - 'npm ci --cache .npm --prefer-offline'
+        build:
+            commands:
+              - 'if [[ "$AWS_BRANCH" == "main" ]]; then npm run build:staging; fi'
+              - 'if [[ "$AWS_BRANCH" == release* ]]; then npm run build:production; fi'
+              - 'if [[ "$AWS_BRANCH" != "main" && "$AWS_BRANCH" != release* ]]; then npm run build; fi'
```
rubo, line 11: By default, it should build for staging and only on `release*` branches for production:

```suggestion
              - '[[ "$AWS_BRANCH" == release* ]] && npm run build || npm run build:stage'
```

package.json
rubo, line 1: These changes must be reverted.

.env.staging
rubo, line 1: This file must be removed.

rubo (changes requested):
As I mentioned earlier, there's #196 for separating the staging and production environments. There's no need to duplicate changes in another PR. If you want to complement #196 with other changes, this PR should target the branch of #196.
With this said, I'm changing the base branch to `feature/env-stage`. We need only `amplify.yml`.

PR #203, opened 2024-08-30
Add promoted operators

vincenthongzy:
Two pages with slightly different information displayed:
1. `/operators`
2. `/avs/:address/operators`

Some edge cases to consider:
1. When we search for operators, we should always display the ads even if the result is empty.
2. In `/avs/:address/operators`, it is ok to not show promoted operators if they are not currently securing the AVS
3. If the promoted operators endpoint fails, we don't need to display error.

Backend PR:
https://github.com/NethermindEth/restaking-dashboard-backend/pull/114

### DIFF
--- a/src/avs/AVSDetailsOperatorsTab.jsx
+++ b/src/avs/AVSDetailsOperatorsTab.jsx
@@ -252,7 +252,7 @@ function AVSOperatorsList({ address, avsError, isAVSLoading, tvl }) {
   const renderPromotedOperators = useCallback(() => {
     return state.promotedOperators?.map((operator, i) => (
       <TableRow
-        className="max-h-16 min-h-16 cursor-pointer border-t border-outline bg-black/100 transition-colors hover:bg-default"
+        className="size-16 cursor-pointer border-t border-outline bg-black/100 transition-colors hover:bg-default"
         key={`promoted-operator-item-${i}`}
         onClick={() =>
           navigate(`/operators/${operator.address}`, {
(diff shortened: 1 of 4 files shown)


Zishan-7 (changes requested):
On `/operators` page when search result is empty, the promoted operators rows expand to take all of the available space

Also ETH -> USD conversion shows `NaN` when search result is empty. This happens on the AVS details page as well

Screenshot 2024-09-03 at 12 34 39 PM

vincenthongzy:
> On `/operators` page when search result is empty, the promoted operators rows expand to take all of the available space
> 
> Also ETH -> USD conversion shows `NaN` when search result is empty. This happens on the AVS details page as well
> 
> Screenshot 2024-09-03 at 12 34 39 PM

Hm strange, the expanding rows on empty results should have been fixed

vincenthongzy:
> On `/operators` page when search result is empty, the promoted operators rows expand to take all of the available space
> 
> Also ETH -> USD conversion shows `NaN` when search result is empty. This happens on the AVS details page as well
> 
> Screenshot 2024-09-03 at 12 34 39 PM

Fixed by decoupling the promoted operators' rate and normal operators' rate on both BE and FE

src/avs/AVSDetailsOperatorsTab.jsx
```diff
@@ -210,6 +213,20 @@ function AVSOperatorsList({ address, avsError, isAVSLoading, tvl }) {
...
+  useEffect(() => {
+    (async () => {
+      try {
+        const response = await rdService.getAVSPromotedOperators(address);
+        dispatch({
+          promotedOperators: response.results,
+          promotedOperatorsRate: response.rate
+        });
+      } catch (e) {
+        //TODO: handle error
```
rubo, line 225: Why no error handling?

vincenthongzy: If the promoted operators endpoint fails, we don't need to display error, so we just swallow it up

Zishan-7 (approved):

src/avs/AVSDetailsOperatorsTab.jsx
rubo: That `TODO` confused me.

rubo (approved):

src/avs/AVSDetailsOperatorsTab.jsx
vincenthongzy: I'll reword this comment

PR #207, opened 2024-09-10
Integrate Clerk with backend services

rubo:
Updated backend services with Clerk session data. They now send the session token automatically if any. The UI has been modified so that it does not render while the Clerk is initializing to avoid sending irrelevant info to the server.

Requires https://github.com/NethermindEth/restaking-dashboard-backend/pull/118

### DIFF
--- a/src/@services/ServiceContext.jsx
+++ b/src/@services/ServiceContext.jsx
@@ -1,9 +1,10 @@
-import { createContext, useContext } from 'react';
+import { createContext, useContext, useEffect } from 'react';
 import AVSService from './AVSService';
 import EigenLayerService from './EigenLayerService';
 import LRTService from './LRTService';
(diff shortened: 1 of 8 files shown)


Zishan-7 (approved):

src/@services/ServiceContext.jsx
```diff
@@ -1,27 +1,35 @@
...
+
+  useEffect(() => {
+    context.session = session;
+  }, [session]);
+
   return (
     <ServiceContext.Provider value={services}>
       {children}
     </ServiceContext.Provider>
   );
```
vincenthongzy, line 25: How often does `session` change? Wouldn't this cause excessive re-rendering of the **WHOLE** component tree? 

Maybe it makes more sense to have the individual pages call `useSession` and pass in the context to the service method instead.

PR #224, opened 2024-10-15
Fix promotion operator navigation

Zishan-7:


vincenthongzy (approved):
did this never work? that's my bad

